PR #14, opened 2019-07-17
Updates to README and gitignore

mathomp4:
* Add notes about NAS and GMAO Desktops for `module use`.
* The `.gitignore` should ignore build and install dirs of any name (mainly the `Debug` ones from `parallel_build.csh`


mathomp4:
Added some changes to `parallel_build.csh` as well. The Debug switch was broken

PR #58, opened 2021-03-03
Bugfix/msienkie/#49 sp interp crash

gmao-msienkie:
This modification was implemented in MERRA2 but was not included in the tag brought over from CVS.
It needs to be implemented to avoid a crash in the 'prevents.x' routine when running with input files produced after Jack's change to prepdata that was implemented in late October 2020.


rtodling:
Meta, can you change this PR so that the request if for a merge to develop instead of main? 
I learned from Matt: next to the PR title (on the write) there is an edit button, if you click on that it will show an option that allows you to redirect the PR to another branch. 
Thanks.
I think the workflow should be user -> develop -> main

mathomp4:
@rtodling I actually did it for her. I wanted to make sure things didn't break after I changed the default. @gmao-msienkie can you review the  changes to make sure they are still what you expect.

gmao-msienkie:
It looks okay to me.  The changes are in only one file – NCEP_Paqc/GMAO_prev/gblevents_gmao.f

rtodling (approved):

PR #60, opened 2021-03-04
line break fixes

rtodling:
This fixes line breaks in the script g5asens.j that only affects developers use of ADAS - it does not affect OPS mode of running. Can we bring this up to main, please?

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,4 +121,5 @@ cd install/bin
 
 ### Run the ADAS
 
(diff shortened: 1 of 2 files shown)


rtodling:
Note I also made a change to README.md in the main fixture to point the current Wiki ADAS documentation. Although I am still working on the doc, it is worth having the pointer correct as most common questions are already in the documentation.

mathomp4 (approved):
Approved!

PR #72, opened 2021-03-11
Use CMake to ignore build and install directories

mathomp4:
A la https://github.com/GEOS-ESM/GEOSgcm/pull/258, use CMake itself to ignore build and install directories rather than assuming all build and install directories will be named `build` and `install`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,9 +2,6 @@
 /@cmake/
 /@modules/
 /@env/
-/BUILD/
-/build*/
-/install*/
 /.mepo/
 parallel_build.o*
 log.*
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,3 +44,14 @@ include_directories(${MPI_Fortran_INCLUDE_PATH})
 add_subdirectory (@env)
 add_subdirectory (src)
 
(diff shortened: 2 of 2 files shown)


tclune (approved):

rtodling:
This seems fine to me.

rtodling (approved):

rtodling:
BTW: 

Whatever writes the sentence:
All required statuses and check runs on this pull request must run successfully to enable automatic merging.
Could be revised to read:
All required statuses and check runs on this pull request must run successful to enable automatic merging.

rtodling:
Oh my .. that was a foolish comment ... I misread the line - please forgive me

rtodling:
The question is why does this show: All checks have failed?

tclune:
It shows "all" checks have failed because the 1 check we are requiring has failed.    Something do do with the required-labels.    I can guess that there is something wrong with the implementation of the "Needs Lead Approval" label, as presumably that would mean @rtodling, who has in fact approved.   So back to @mathomp4

mathomp4:
I guess I was leaving the "Needs Lead Approval" removal to @rtodling, but as he approves, I removed it. 

That said, I wonder if there is a way to remove that label on approval...hmm... 🤔

PR #77, opened 2021-03-15
required to be able to launch sscript by hand - from crashed case

rtodling:
need defining EXTDATA env for script to work when submitted by hand (offline)


mathomp4 (approved):
Approving since I trust Ricardo on these scripts!

Issue #79, opened 2021-03-16
coupling LDAS with ADAS


saraqzhang:
@rtodling @gmao-jstassi @mathomp4 @gmao-rreichle 
Modifications and addition in ADAS  scripts  for coupling LDAS with ADAS:
(1) with the option of LDAS-ADAS-coupling  on,  ADAS will run land analysis ( GEOS LDAS).  The resulting land increments  will be applied to AGCM  during the corrector segment.   The diag lfo collection during  the predictor segment will provide atmospheric forcing to LDAS ensemble land model integration. 
(2)  with the option of LDAS-ADAS-coupling off (default),  ADAS produce zero diff results comparing to that of the system before changes are made. 

The changes had been implemented and tested in a sandbox system from the cvs-based ADAS (cvs tag: GEOSadas-5_26_0_p3)
The changes are implemented and tested in feature/saraqzhang/ladasdev.

rtodling:
This is now folded into rt-GEOSadas-5_29_3 the upcoming v5.29.3 release

PR #80, opened 2021-03-16
       add option for coupling LDAS with ADAS

saraqzhang:
Addresses issue #79 -- integration of LDAS into ADAS.

0-diff tested by @saraqzhang 

Because the ADAS does not yet use the MAPL2 version of the GCM, this branch does **not** yet run with LADAS unless temporary modifications are made to include a GCM_GridComp and GCM_App that are running with MAPL1 but include functionality for ingesting LDAS increments into the GCM.  This is left open for now. 

	modified:   src/Applications/GEOSdas_App/GEOSdas.csm
	modified:   src/Applications/GEOSdas_App/fvpsas
	modified:   src/Applications/GEOSdas_App/fvsetup
	modified:   src/Applications/NCEP_Etc/NCEP_enkf/scripts/gmao/atm_ens.j
	new file:   src/Applications/NCEP_Etc/NCEP_enkf/scripts/gmao/atmos_eldas.csh
	modified:   src/Applications/NCEP_Etc/NCEP_enkf/scripts/gmao/etc/AGCM.rc.tmpl
	modified:   src/Applications/NCEP_Etc/NCEP_enkf/scripts/gmao/etc/HISTAENS.rc.tmpl

Flowchart of LADAS implementation as of 5 Oct 2020 (pdf):
[ladas-dev-schemtic2.pdf](https://github.com/GEOS-ESM/GEOSadas/files/7292793/ladas-dev-schemtic2.pdf)
:


src/Applications/GEOSdas_App/GEOSdas.csm
```diff
@@ -3189,53 +3198,100 @@ endif
...
+  Sub StageLdasIncr_()
+#-----------------------
+   if ( $?ECHO___ ) set echo
+
+    cd $FVWORK
+
+    set LINC_DIR=$LDHOME/output/*/cat/ens_avg/
+
+    # default: 
+    set ldas_int = 10800
```
gmao-rreichle, line 3252: Should we really hard-wire the LDAS and ADAS time steps here?  Do these numbers exist in some variables somewhere?

gmao-rreichle:
Thanks, @saraqzhang , for putting the PR together. 
I don't know anything about the ADAS scripts, so I can't add much in terms of review.
I added one inline comment/question about hard-coding time steps, see above.
Also, it looks like the indentation is off across much of the new code, which should be fixed before the merge.
Finally, I added the "Do Not Approve" label (per the process used by the GCM and GEOSldas groups) to express the fact that this PR isn't ready for formal approval because we're still planning to meet with @rtodling and @gmao-jstassi to discuss the changes.

rtodling:
Sara,

As you know there is a lot in your PR. So I am going to comment for now only on GEOSdas.csm. 

I understand why you made changes as you did - since the original implementation of Joe's was rather intrusive to the CSM. I think there is a different, much less intrusive and preferred way of making changes to the CSM.

It seems to me the main routines were placed in the CSM by Joe - and you edited them according to your needs. What I would suggest is for example, something like this:

1) instead of making the changes as you made in the subroutine LandAnalysisRun_, what I would do is, 
1a) go back to the original CSM; wipe out the contents inside the subroutine; write a simple call to an external csh script - e.g., land_analysis_run.csh
1b) fill in the new script with the contents you have in the CSM as in submitted in this PR. 
1c) the subroutine you create, e.g., land_analysis_run.csh should have a very simple interface ... pass date/time and pretty much only that. All other variables will be a part of the env - including FVWOK, FVROOT, EXPID, so you don't need to pass any of that. This should hit any complexities from the CSM.

This way, instead of adding complexity to the CSM, you would be cleaning it up. Also, this gives you much more flexibility and control over what you put inside your scripts; you'll be able to change it w/o having to coordinate w/ changes in the CSM unreeled to yours; this is separation of interests.

2) I would also say in passing that marking codes w/ thing like:

#sqz--merge--start
#sqz--merge--end

3) Things that are outside main existing LDAS subroutines in the SCM, such as the following chunk of code 

     # lfo files held from last segment for ldas coupling 
     #--------------------------------------------------- 
      if( $LDAS_ANA ) then
      /bin/mv $FVHOME/recycle/holdforc/*2d_lfo*nc4  $FVWORK/.
      endif

should actually be move to inside your scripts (such as the one referred to in 1 above). It's possible that a couple of these might still need to be left as you coded - that is scattered little if's here and there that take care of something small. I don't think the one above is a case in point, but perhaps, something like

#sqz--merge--start
        # hold lfo files for the next segment of ldas coupling
        #----------------------------------------------------------
           if  ( $LDAS_ANA ) then
           mkdir -p       $FVHOME/recycle/holdforc
          /bin/cp  *2d_lfo*nc4  $FVHOME/recycle/holdforc/.
           endif
#---------

might need to stay as is - w/o the personal change markers (see 2 above).

What I suggest here for GEOSdas.csm is similar for how things are done for the ensemble component. The ensemble, as I wrote originally - and you followed well - has lots of separation of interest ... following the suggestion above would do the same when it comes to the central DAS.

saraqzhang:
@rtodling @gmao-rreichle 
Ricardo,
I see. You recommend that for an implementation of this "separate interest" component in the central DAS, it should take the approach similar to that in the ensemble DAS (like atmos.eldas.sch).  I will follow the recommendation, make a new script and cleanup the original GEOSdas.csm.  

Sara

saraqzhang:
@rtodling @gmao-rreichle 
Ricardo, 
I revised the implementation in GEOSdas.csm and fvpsas following your recommendation and is making test runs. I plan to push the changes for your review after the test results are verified.  However, before I push the new changes to this branch, I would like to push a tag for the branch at the current instance, since this is the system used in previous coupling das experiments with results verified.  I pushed a tag at  external GEOSgcm_GridComp The tag would be something like "ladas_coupling_cvs5271_v1.0". Let me know if you have suggestion on the tag name.   
I have a branch  at  the external GEOSgcm_GridComp  that contains changes for ldas increment application, and I pushed a tag there for that branch as well. 

Sara

saraqzhang:
@rtodling @gmao-rreichle 
Ricardo,  
I will not create a tag  for the snapshot of the current branch as I proposed in the last comment.  I will use commit record to keep track the versions and changes within the branch. 

Sara

saraqzhang:
@rtodling @gmao-rreichle 
Ricardo: 
I committed the changes to the scripts (d9a2272, 13ffad8 ).  Following your suggestions, ldasRun function is removed from GEOSdas.csm. the ldas coupling is handled  by an external script.  The run test with LDAS_COUP off verified Zero-diff to git adas develop.  The run test with LDAS_COUP on verified  Zero-diff to the version before the changes. 

Sara

gmao-rreichle:
Thanks, @saraqzhang.   A couple of thoughts:
1) When I view the changes in Github, the indentation seems to be a bit uneven, more so than what's in the existing portions of the scripts.  It's not for me to decide, but clean indentation would probably be helpful. 
2) There's a good bit of overlap in how the LDAS scripts for the central member (`ldas_run.csh`) and the atm. ensemble (`atmos_eldas.csh`) work.  Is there any use in merging these scripts into a single script that can execute each of the necessary functions?  Having one script would avoid having to maintain similar code in two places, but it may also get too complex.  This may or may not be a good idea when we look ahead to a tighter integration of the LDAS and atm. ensembles.  Just a thought.  @rtodling : Do you have a preference either way?

saraqzhang:
Ricardo, 
I committed a few more changes to the scripts under review in this PR, mostly in the external scripts ( ldas_run.csh and atmos_eldas.csh ) to incorporate information from a new version of LDAS.rc in GEOSldas.   It passed 0-diff test. 

Sara 

@rtodling @gmao-rreichle

src/Applications/GEOSdas_App/GEOSdas.csm
```diff
@@ -5695,6 +5646,13 @@ endif
               /bin/cp $EXPID.ana_radstat_rst.$rtag.tar $RSTHOLD/ &
             endif
 
+        # hold lfo files for the next segment of ldas coupling
+        #----------------------------------------------------------
+           if  ( $LDAS_ANA ) then
```
gmao-rreichle, line 5651: It looks like the indentation here is messed up whereas it looks ok in the surrounding code.  This should be fixed.

src/Applications/GEOSdas_App/fvpsas
```diff
@@ -474,6 +493,22 @@
...
          endif
 
+#------------------------------------------------
+#  check ldas job status and stage ldas increments 
+#--------------------------------------------------
+       if ( ( $LDAS_ANA ) && ( $LDASFDBK ) ) then
+        echo " LDAS coupling: stage 1"
+                                                      zeit_ci.x ldasStage
+       ldas_run.csh 1 060000 |& tee -a ldasrun.log
+      if( $status) then
```
gmao-rreichle, line 503: indentation?

src/Applications/GEOSdas_App/fvsetup
```diff
@@ -538,6 +538,20 @@ $ENV{"PATH"} = $FindBin::Bin .":$ENV{PATH}";
...
+    print "\n checkprintout  ldas_ana == 1 ($ldas_ana)\n";
+    if ($ldas_ana == 1 ) {
+    print "\n Please run ldas_setup following instructions in GEOSldas/README.md\n";
+    print "\n requires config input files: YOUR_exeinp.txt, YOUR_batinp.txt.\n" ;
+    print "\n example LDROOT: /discover/nobackup/$user/GEOSldas\n" ;
+    print "\n example LDHOME: /discover/nobackup/$user/ldascoup_C180\n" ;
+    print "\n example cd $LDROOT/install/bin/ \n";
+    print "\n example salloc \n";
+    print "\n example source g5_modules \n" ;
+    print "\n example ldas_setup setup $LDHOME exp_exeinp.txt exp_batinp.txt \n" ;
```
gmao-rreichle, line 552: Do we really need these lines ?  This should be covered in `GEOSldas/README.md`.
Not sure about `salloc` command.
But most importantly, does the GEOSldas experiment need to be set up in a specific directory, or is there a place where we can tell the ADAS where the GEOSldas experiment is set up?  If the "example" directories have to be followed to the letter, we would need to say so here.

src/Applications/GEOSdas_App/ldas_add2rc
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+MYNAME="ldas_add2rc"
```
gmao-rreichle, line 3: A different name for this script might be more informative.  Maybe `reconcile_ldas_adas_rc`?

src/Applications/GEOSdas_App/ldas_add2rc
```diff
@@ -0,0 +1,71 @@
...
+fi
+
+FVHOME="$1" 
+LDHOME="$2" 
+
+cd $FVHOME/run
+cp AGCM.rc.tmpl  AGCM.rc.tmpl.original 
+
+list="LANDASSIM_DT LANDASSIM_T0 "
+  for var in $list
```
gmao-rreichle, line 37: indentation seems off throughout this script

src/Applications/GEOSdas_App/ldas_add2rc
```diff
@@ -0,0 +1,77 @@
...
+
+sed -i '/LDAS_INCR/r add.txt' AGCM.rc.tmpl
+
+echo " The following land parameters are added to AGCM.rc: " 
+for var in $list
+do
+echo "$(grep $var add.txt)"
+done
+
+# add same resource parameters into AGCM.rc for atm ens
```
gmao-rreichle, line 50: I moved up the block that adds LANDASSIM_DT and LANDASSIM_T0 into AGCM.rc so it's together with the same action for the central AGCM.rc.  Not sure if the logic is still intact.
Also, what happens if the ADAS is run in 3D-Var configuration?  In this case, the ./atmens/ directory should not exist.  This script might need to know if the ADAS is in 3D-Var or in Hybrid 4D-EnVar and do its thing accordingly.

src/Applications/GEOSdas_App/ldas_add2rc
```diff
@@ -0,0 +1,77 @@
...
+Lset="$(grep $var $LDHOME/run/LDAS.rc)"
+Aset="$(grep $var $FVHOME/run/AGCM.rc.tmpl)"
+lv="$(echo $Lset | cut -d " " -f 2)"
+av="$(echo $Aset | cut -d " " -f 2)"
+if [ "$lv" != "$av" ]; then
+echo " " 
+echo "Please repeat setup such that the following parameter values are consistent in AGCM.rc and LDAS.rc:"
+echo "in AGCM.rc: $Aset "  
+echo "in LDAS.rc: $Lset "
+echo " "
```
gmao-rreichle, line 74: This script should halt the LADAS setup if the LDAS.rc and ADAS.rc have inconsistent values.  I don't see this "exit" or "stop" condition, but perhaps I'm missing something.

src/Applications/GEOSdas_App/ldas_run.csh
```diff
@@ -0,0 +1,206 @@
...
+  env
+  echo " ${MYNAME}: not all required env vars defined"
+  exit 1
+endif
+
+set stage = $1
+set freqa = $2
+
+
+     cd $FVWORK
```
gmao-rreichle, line 59: indentation?  (throughout the script)

gmao-rreichle:
@saraqzhang, 
I added a couple of commits that mainly clean up the comments/help text, along with a few inline comments (see above).  I hope I didn't break anything.

src/Applications/GEOSdas_App/GEOSdas.csm
saraqzhang: @gmao-rreichle 
I will make it a task to comply with the indentation convention and do one commit for all.

src/Applications/GEOSdas_App/ldas_add2rc
saraqzhang: Maybe we move "rc" to after "reconcile",  a name ending with  "rc" could be misread as a rc file.

src/Applications/GEOSdas_App/ldas_add2rc
saraqzhang: This script will be run after fvsetup and ldas_setup procedures are done. Both adas exp and ldas exp dir are setup with LDAS.rc and AGCM.rc.    If the script finds inconsistent values,  all it can do  is to print warning message. We wanted to check all parameters in the list ( not stop and exit at the first found inconsistency),  then it reaches  the end,

src/Applications/GEOSdas_App/fvsetup
saraqzhang: LDROOT and LDHOME are needed at the time of fvsetup ( input to fvsetup).   The rest of the lines are not essential  and can be removed. 
 The path and names of "GESOldas"  and "ldascoup_C180" are just example, not to be followed.

src/Applications/GEOSdas_App/ldas_add2rc
saraqzhang: This script is run after fvsetup and ldas-setup ( so that AGCM.rc and LDAS.rc are in place in exp/run ) .  It should go through the given list of parameters and prints out warning message if any one is inconsistent,  and not exit till reaching the end of the parameter list.

src/Applications/GEOSdas_App/ldas_add2rc
saraqzhang: the original sequence ( atmens at the last step) would exit however not gracefully when in case of 3D-var,  while for the central all checking/adding are already carried out.   I will add "3dvar" to the argument input, so when that is present the script should skip atmens part.

saraqzhang (commented):
@gmao-rreichle 
I put a few responding comments following the thread of your comments and commits.

src/Applications/GEOSdas_App/ldas_add2rc
gmao-rreichle: maybe `reconcile_ldas_adas_config` then?

src/Applications/GEOSdas_App/ldas_add2rc
gmao-rreichle: I don't think we should have non-graceful exits, so we need to have an "if" statement for presence/absence of the atm ensemble.  
It makes sense to run through all parameter values and tabulate the ones that are inconsistent before exiting.
But what still seems to be missing is a "return code" and controlled exit.  This script should return something to the main script that tells the main script to exit with a message.  If the list of inconsistent parameters is written by this script, then the main script can simply exit with a brief error message, something like "ldas_add2rc: inconsistent ADAS and LDAS config"

src/Applications/GEOSdas_App/ldas_add2rc
gmao-rreichle: It makes sense to run through all parameter values and tabulate the ones that are inconsistent before exiting.
But what still seems to be missing is a "return code" and controlled exit.  This script should return something to the main script that tells the main script to exit/stop with a message.  If the list of inconsistent parameters is written by this script, then the main script can simply exit with a brief error message, something like "ldas_add2rc: inconsistent ADAS and LDAS config"

src/Applications/GEOSdas_App/fvsetup
gmao-rreichle: It doesn't seem to me that the sample directories for LDROOT and LDHOME are of much help here, so I would delete these lines as well.  Rather, the key message here should be that LDROOT and LDHOME need to be specified consistently in fvsetup and ldas_setup.  Can this message about ldas_setup be placed such that LDROOT and LDHOME are already known to fvsetup?  Then the message could print LDROOT and LDHOME and tell the user to specify these values in ldas_setup.  
If LDROOT and LDHOME end up in a config file somewhere, maybe we could also apply the "ldas_add2rc" consistency check to LDROOT and LDHOME?
Perhaps I'm missing something, but I have a sense that we can still make the interplay of fvsetup and ldas_setup more robust.

src/Applications/GEOSdas_App/ldas_add2rc
saraqzhang: This script is not called by any main script.  It is a standalone "post setup check"  after fvsetup and ldas_setup are finished.

src/Applications/GEOSdas_App/ldas_add2rc
saraqzhang: yes i will add 3dvar / hybrid  case choice  and the corresponding "if"  block.

src/Applications/GEOSdas_App/fvsetup
saraqzhang: at the time of fvsetup , ldas_setup is not yet run.   User has to have decision before run fvsetup on where his ldas root and  ldas exp home would be and give those info to fvsetup. ( and he should keep the same path/name  later when do ldas_setup ).   LDROOT and LDHOME setup by fvsetup  will end up in adas  top run script and put into env. variables during the run.

gmao-rreichle: Thanks for the clarification, and sorry I'm so slow to catch up.  From what I understand now, though, it looks like we could indeed **print this message about ldas_setup _after_ fvsetup has been given LDROOT and LDHOME**.  Then the message could echo the LDROOT and LDHOME values and tell the user that the exact same values need to be specified (under their GEOSldas names) in ldas_setup.  Does that make sense?

src/Applications/GEOSdas_App/ldas_add2rc
gmao-rreichle: Thanks, I see.  I'm still not sure how best to sort out this chicken-and-egg scenario.  Could we possibly have fvsetup call ldas_setup?  
If I understand the discussion correctly, fvsetup is provided with LDROOT, so it should be able to run ldas_setup.  One challenge is that ldas_setup requires the LDAS "exeinp" and "batinp" config files, which would then also have to be provided to fvsetup.  But the user has to have these files anyway, so providing them to fvsetup doesn't really change anything.  
Another complication is that g5_modules is probably different between the ADAS and the LDAS.  If ldas_setup is called at the very end of fvsetup, could we have fvsetup source (LDAS) g5_modules just before calling ldas_setup? 
If that's possible, then fvsetup could also call ldas_add2rc after calling ldas_setup. 
Alternatively, maybe a new main script, fvsetup_ladas, could be added to the ADAS repository.  This script would first find out LDROOT, LDHOME, LDAS "exeinp"/"batinp", then call fvsetup, then call ldas_setup, and finally ldas_add2rc.
Maybe I'm too optimistic about constructing a single master script that sets up both the ADAS and the LDAS, but if the LADAS is to be used more broadly we have to make the setup as easy and as robust as possible.
For now, we can wait until Ricardo has had a chance to take a closer look.  I imagine he will have some thoughts on this as well.

saraqzhang:
@rtodling  
Ricardo,
to bring the ldas coupling development  ( this PR)  to sync with the updates of adas, we merged the develop branch ( as 7/2/2021)  into the branch where this PR is. 
In the process of testing hybrid adas ( use the develop pulled as is, not the merged branch)  we encountered the problem of  missing some run-time lib in atm_ens. I sent you and Ben an email today about it. Hope it can be resolved soon with your help,  so that we can complete the test for the control and move on to test the merged system with  ldas coupling. 

Sara 

@gmao-rreichle

saraqzhang:
@rtodling 
Ricardo,
I just added ldas_run.csh to GEOSdas_App/CMakeList.txt so that it can be installed in bin/. 
With the fix you put in atm_ens.j last week,   I have finished the zero-diff test and the  coupled ldas-adas test in  hybrid cycling runs , results are verified. 

Sara

rtodling:
Glad to hear Sarah. I will be looking at your PR next in my list - since I trust your zero-diff test - and should be able to reproduce it myself. I will keep you posted.

tclune (approved):
CMake changes OK.

Sorry for the delay - had a problem with my email filters.

saraqzhang:
@rtodling 
Ricardo,
the update to gcmGridComp ( PR442)  for coupling ldas adas has been approved and merged  in gcm v10.19.3. 
hope  you will have time to review this PR soon and make the merge to adas that is in line with gcm v10.19.3. 

Sara

saraqzhang:
@rtodling 
Ricardo,  I  merged develop branch  ( with Joe's updates recently since my last merge) into the PR#80 branch here to resolve any conflict so far.   We really hope that you can find time to review/merge this PR before the start of  x46 experiment.  -Sara

rtodling:
I might be missing something here ... but it looks to me the variables ldasexp and ldasexp4ens used in fvsetup are not defined anywhere! What are these variable meant to be?

saraqzhang:
during the fvsetup session, if   ldas_ana  is yes (1)     users are prompted  to input these two variables.  if  ldas_ana is no  (0) , these two will be empty.

saraqzhang:
when ldas_ana is 1 ,  ldasexp is  where ldas exp  (coupled with atmos central )directory is, and  ldasexp4ens is where ldas exp  (coupled with atmens  )directory is.

rtodling:
What exactly do these correspond to?  Besides ... I don't think this will work as you say. The line:

For example: 
The first instance of the variable $ldasexp  is in the query below:
$ans3 = query("LDAS HOME = /discover/nobackup/$user/$ldasexp, full path?  ");
The query prompts the user showing a default which should be a completely resolvable variable - which the above is not since nowhere a default is set.

This would work:
$ldasexp = something  # (I don't know if $ldasexp  is a name or a directory or what ... but anyway 
$ans3 = query("LDAS HOME = /discover/nobackup/$user/$ldasexp, full path?  ");

Depending on how the user answers the query you'd have the answer to be either
/discover/nobackup/rtodling/something
or whatever the user types at command line
something_I_type

Otherwise, perl will complain about an undefined var!

saraqzhang:
sure I  will make a note on this, put something default there.

rtodling:
Pls Sara: what is the meaning of these variables?

rtodling:
I think this is asking too much from a user ...

The question is this: why is LDHOME any different than FVHOME? 
and why is LDHOME4ens any different than FVHOME/ATMENS ... or is this what it is typically ... 

Also, I will change the env var name LDHOME4ens to LDHOME4ENS  (convention is for env var to be upper case); but pls be patient w/ me and give me a little explanation of the flow ...

rtodling:
I don't know why but some of the scripts have considerable indentation issues ... I am fixing them all

saraqzhang:
it is harder to put LDAS in FVHOME at the current implementation,  for LDAS  is built separately from adas and has its own setup procedure and  layered directory structures.  Maybe down the road the coupling can be more integrated.

rtodling:
I think at some point you emailed me a workflow, didn't you? Can you email that to me again, pls? thanks

saraqzhang:
It will be better if we can implement these two in the config.. when we put them in the job.j scripts we tried  to keep the intrusion to  adas various files as limited  as possible. 
I will send you the workflow chart tonight.

rtodling:
What is this comment/printout in fvsetup:

  print "\n example LDROOT: /discover/nobackup/$user/GEOSldas\n" ;

Nowhere do I see a variable called LDROOT ... which - according to your flowchart is something I was expecting to see somewhere! But maybe it's not needed since everything is hidden in LDHOME4ENS/run/lenkf.j  - I don't know since I don't have access to this.

BTW: The script atmos_eldas.csh refers to a variable named:  $PBS_BIN that does not exist in SLURM - I don't know what that's needed ... this is what happens when the code tries to execute this line in batch:

set jobIDlong = `$PBS_BIN/sbatch $jobldas`
PBS_BIN: Undefined variable.
set jobID = `echo $jobIDlong  |awk -F'[ ]' '{print $4}'`
awk -F[ ] {print $4}

I ran just those lines in a separate script. This needs a fix.

I am putting one that simply removes $PBS_BIN/ and leave on sbatch ... but you should corroborate that's good for you.


Also, I changed the default in the queries in fvsetup related to your 2 env vars to something that at least makes sense - and it doesn't refer to an undefined variable:

LDAS HOME = /discover/nobackup/rtodling/ldas01, full path?   
LDAS HOME for land ensemble = /discover/nobackup/rtodling/ldas01/run/atmens, full path?

(two lines above are the echo from fvsetup of the defaults now)

rtodling:
I'm also going to make a general comment: whatever edit you are using to edit the scripts/files is messing up dramatically the indentation in the files.  I ask that you please be cognizant of that.

saraqzhang:
PBS_BIN  is a leftover from an old script,  it should be removed.   I just found a bug in ldas_run.csh as well ( another leftover from old script ) .   we will exam and edit these scripts in review meeting this afternoon.

saraqzhang:
@rtodling    I just pushed a modified ldas_run.csh  to this branch: removed undefined env variables, saved ldas job id in txt file.  The changes are tested in a coupled exp run.

saraqzhang:
yes  the worst case would be that.  This  modification does not impact any zero-diff results. It fixes the job monitoring problem in LADAS run particularly in the situation that ldas takes longer time to finish. 
leave it to Ricardo, he can ignore this update when merging his branch to develop.

gmao-rreichle:
Addressed with PR #121.

PR #89, opened 2021-04-30
Update README with more git clone info

mathomp4:
I've had some users encounter the "Permission denied (publickey)" error with the SSH URL, so this is an attempt to add a bit more specificity to cloning GitHub repos. Since users might see that GitHub offers more than one way to get a repo (HTTPS, SSH, GitHub CLI), I stayed on the side of completeness and talked about all three.

If preferred, I can just stick with the instructions for SSH, but users will always find every way to clone thanks to Google 😄


rtodling (approved):
This seems fine to me; you know better.

PR #92, opened 2021-05-25
Feature/msienkie/aircraft bias updates

gmao-msienkie:
The "bugfix" is with respect to compiling "cft_update.f" - it needs to be compiled with the convert little endian flag or the program won't run at all.

There are some additions on this branch - a newer version of cft_update.f with an additional option.  The program is otherwise zero diff in calculating the offline aircraft bias correction.

There is also an added program 'tailwrite.f' that I use in splitting up the bias coefficients by tail number for monitoring and comparison of the bias correction coefficients with e.g. the variational aircraft bias correction or with the bias correction coefficients from other experiments or from MERRA2.


gmao-msienkie:
the components.yaml does not need to be included in this merge - I just picked it up by accident.

rtodling:
Meta, can you please explain this to me? What does this mean? I have been running x0044 from the GIT compiled version of the source code. Does this mean the aircraft bias correction has not been running correctly in x0044?

gmao-msienkie:
This code is for the external aircraft bias correction that we are using in MERRA2.  It is used if 'setenv ACFTBIAS 1' is in the job script.
By default when the aircraft bias correction option is chosen in recent tags, 'setenv ACFTBIAS 2' is added to the job script.
So the issue here is for anyone who wants to 'run like MERRA2' using the Git code.

rtodling:
Meta can you please work on the conflicts for this PR so I can approve it?

gmao-msienkie:
I brought in the 'components.yaml' changes from 'develop' so the conflict should be resolved.

mathomp4 (approved):
Approved per @rtodling approval on email 😄

PR #95, opened 2021-06-03
Feature/rtodling/geo sadas 5 28 0

rtodling:
Bringing all that now supports x0045 to develop.

PR #96, opened 2021-06-03
Update MOM5 to use relative remote

mathomp4:
This one slipped by me. The MOM bit of `components.yaml` was using a hard-coded SSH URL which breaks the use of HTTPS urls.


rtodling (approved):
This is now on the branch associated w/ PR #95 ... it will go into develop when that merge takes place

PR #107, opened 2021-07-26
Feature/jstassi/#73 integrate radmon into adas 2

gmao-jstassi:
Updates for integrating RADMON into ADAS

### DIFF
--- a/src/Applications/GEOSdas_App/GEOSdas.csm
+++ b/src/Applications/GEOSdas_App/GEOSdas.csm
@@ -140,6 +140,7 @@
   if ( !($?LOCAL_ACQUIRE) )       setenv LOCAL_ACQUIRE 0
   if ( !($?MAP05RST) )            setenv MAP05RST      0
   if ( !($?MKRESRST ) )           setenv MKRESRST      0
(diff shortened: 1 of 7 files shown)


gmao-msienkie:
I just wanted to mention that, rather than processing the Pyradmon gsidiag_bin2txt a month (or partial month) at a time, I think it would be more convenient to have people create the diag.txt files when running each job segment.  There have been hooks for this for quite a while in the DAS scripts for running the gsidiag_bin2txt and there are listings in the silo.arc to copy the text files from $FVWORK.
To activate this, you would just need to add one line to the FVDAS_Run_Config with the command to run gsidiag_bin2txt

setenv GSI_DIAG2TXT "gsidiag_bin2txt.x -passivebc -npred 12"   (or something like that with the appropriate options)

rtodling:
Joe, I cannot fold this into develop - ie approve the PR - since echorc.pl is missing. I imagine you placed it in GMAO_Shared/GMAO_etc ... but I don't see it in a PR under GMAO_Shared - you'll need to do that before this can be approved or else the DAS will break.

rtodling (approved):
I haven't tested myself, but it looks fine - and I trust your testing.

PR #103, opened 2021-07-06
Update components to be in line with GCM

mathomp4:
This PR is to bring the `components.yaml` in the GEOSadas "up to date" with the GEOSgcm after the v5.29.0 release. 

The changes are:

* ESMA_env v3.2.1 → v3.3.0
* ESMA_cmake v3.4.2 → v3.5.2
* MAPL v2.6.7 → v2.7.0
* FVdycoreCubed_GridComp v1.2.13 → v1.2.15
* GEOSgcm_App v1.4.1 → v1.5.2
* GEOSana_GridComp v1.4.1 → v1.4.2

Pretty much all these changes are due to changes in Baselibs from 6.1.0 to 6.2.4 (ESMA_env and ESMA_cmake) to support development with GFE libraries (CMake change), as well as to support newer MAPL (which in `develop` and later tags will ***REQUIRE*** Baselibs 6.2.4).

But, one more reason to facilitate this is because if we can move the GEOSadas up to "in sync" with the GEOSgcm, I can update the CI in MAPL so that *every* pull request to MAPL *must* build the GEOSadas. But, we need to bring GEOSadas up to date.

This should be zero-diff as far as I know because (as @sdrabenh can attest) the switch to Baselibs 6.2.4 was a zero-diff change (v10.19.1 → v10.19.2).

PR #108, opened 2021-07-29
Update GEOSadas to be in line with GEOSgcm v10.19.3

mathomp4:
In the time since #103 was made, the GCM updated to v10.19.3. This PR would update the GEOSadas as well. The updates are:

[GMAO_Shared v1.4.6](https://github.com/GEOS-ESM/GMAO_Shared/releases/tag/v1.4.6)
[MAPL v2.8.0](https://github.com/GEOS-ESM/MAPL/releases/tag/v2.8.0)
[GEOSgcm_GridComp v1.12.4](https://github.com/GEOS-ESM/GEOSgcm_GridComp/releases/tag/v1.12.4)
[fvdycore v1.1.7](https://github.com/GEOS-ESM/GFDL_atmos_cubed_sphere/releases/tag/geos%2Fv1.1.7)
[GEOSgcm_App v1.5.3](https://github.com/GEOS-ESM/GEOSgcm_App/releases/tag/v1.5.3)

Note that this is potentially non-zero-diff. Per the GEOSgcm release notes:

> Potential Non-0-diff Change:
> 1. Upgrade to MAPL v2.8.0. For the MERRA2 GOCART Emissions, all testing shows it is zero-diff. But for the Ops GOCART Emissions, it there are very small roundoff differences. The results are non-zero-diff due to a bug fix (to a race condition) in this version of MAPL on how grids are handled.

@bena-nasa can explain the MAPL bug fix better than I can if needed.

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -28,13 +28,13 @@ NCEP_Shared:
 GMAO_Shared:
   local: ./src/Shared/@GMAO_Shared
   remote: ../GMAO_Shared.git
-  tag: v1.4.5
+  tag: v1.4.6
   develop: main
 
 MAPL:
   local: ./src/Shared/@MAPL
   remote: ../MAPL.git
-  tag: v2.7.0
+  tag: v2.8.0
   develop: develop
 
 FMS:
@@ -52,7 +52,7 @@ GEOSana_GridComp:
 GEOSgcm_GridComp:
   local: ./src/Components/@GEOSgcm_GridComp
   remote: ../GEOSgcm_GridComp.git
-  tag: v1.12.3
+  tag: v1.12.4
   sparse: ./config/GEOSgcm_GridComp.sparse
   develop: develop
 
@@ -77,7 +77,7 @@ FVdycoreCubed_GridComp:
 fvdycore:
   local: ./src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSsuperdyn_GridComp/@FVdycoreCubed_GridComp/@fvdycore
   remote: ../GFDL_atmos_cubed_sphere.git
-  tag: geos/v1.1.6
+  tag: geos/v1.1.7
   develop: geos/develop
 
 GEOSchem_GridComp:
@@ -121,7 +121,7 @@ mom6:
 GEOSgcm_App:
   local: ./src/Applications/@GEOSgcm_App
   remote: ../GEOSgcm_App.git
-  tag: v1.5.2
+  tag: v1.5.3
   develop: develop
 
 UMD_Etc:


bena-nasa:
Yes, as Matt said with the ops emissions there was a small non-zero diff. The issue was that I had previously made a change to ExtData that caused the order the variable to be provided get processed to change. Well it turned out there were two emissions files that were identified as having "equal" lat-lon grids but on close inspection they have ever so slightly different lons and lats (i'm talking very small differences). Well, which ever file got opened first, that lat-lon grid won our grid manger so when the next file was opened it saw, I've already created a grid with that factory. So depending on which was opened first you would get slightly different lons and lats. So this change I made caused this order of these to flip during processing and indeed just change the order these two entries were in the ExtData file would change which got used. So when we realized this we tightened up the lat-lon grid factory so inconsistencies would not happen like this again.

rtodling:
Call me crazy ... but I can no longer find the repo for fvdycore! What happened to it? I am trying to see the diff between what I have in develop (geos/v1.1.6) and what you have here ... but I can find the repo!

mathomp4:
> Call me crazy ... but I can no longer find the repo for fvdycore! What happened to it? I am trying to see the diff between what I have in develop (geos/v1.1.6) and what you have here ... but I can find the repo!

Ah yeah, that was one where the actual repo has a *really* different name than what GEOS called it. NOAA calls it "GFDL_atmos_cubed_sphere":

https://github.com/GEOS-ESM/GFDL_atmos_cubed_sphere

So the difference would be:

https://github.com/GEOS-ESM/GFDL_atmos_cubed_sphere/compare/geos/v1.1.6...GEOS-ESM:geos/v1.1.7

The reason for this was documented by @bena-nasa  here: https://github.com/GEOS-ESM/GFDL_atmos_cubed_sphere/pull/43

rtodling:
It's really odd how and why an optional variable needs to be ifdef'ed around it! Anyway, I will assume this is as the change claims - to be zero-diff. Thanks Matt.

rtodling:
Another odd event in which I get:

git checkout v1.1.7
error: pathspec 'v1.1.7' did not match any file(s) known to git

What's the trick here?

mathomp4:
> Another odd event in which I get:
> 
> git checkout v1.1.7
> error: pathspec 'v1.1.7' did not match any file(s) known to git
> 
> What's the trick here?

The tag is actuall `geos/v1.1.7`. We called it that just in case GFDL had an "official" tag of v1.1.7 (since our code isn't exactly theirs).

tclune:
@rtodling   The summary at the top definitely states that the differences are not always 0-diff.  Probably should have made the statements in the reverse order to minimize communication.   And we should have noticed that your Aug. 3rd comment suggest you had missed the non-zero diff aspect.

rtodling:
Now included in GEOSadas-5_29_3:

MAPL v2.8.0
GEOSgcm_GridComp v1.12.4
fvdycore v1.1.7


As for Shared ... I need feature/rtodling/cascase for the DAS; I think the DAS supersede the version pointed to here: v1.4.6

As for GEOSgcm_App: The DAS uses what's under: feature/rtodling/v1_5_4_onlXoffl_flags which (as far as I can tell) supersedes v1.5.3

mathomp4:
@rtodling Indeed, this is probably out of date. I'll close this as your branches are probably above this.

PR #109, opened 2021-08-09
Fixed bug in fvsetup

gmao-jstassi:
Fixed bug in fvsetup where the script was modified to not query for PYRADMON value (but just take default automatically).


rtodling (approved):
Thanks for that Joe.

PR #120, opened 2021-10-27
Properly finalize GEOSgcmPert.x

bena-nasa:
This fixes the random crashing that was being experienced when running this application with the latest tag.


rtodling (approved):
Thanks Ben.

PR #117, opened 2021-10-08
bug fix for ADAS checking on LDAS job

gmao-rreichle:
This pull request fixes a problem with the ADAS checking on LDAS job completion within the LADAS.  The fix was added to PR #80 _after_ @rtodling had already pulled @saraqzhang's branch to his working branch.

The fix in the present pull request was constructed from Sara's recent commit [5fac454 (link to commit with white-space changes hidden)](https://github.com/GEOS-ESM/GEOSadas/pull/80/commits/5fac4547a982a5e8c032521774fd76c2c2de76c0?w=1), which included a couple of other, trivial changes.

The target branch of the present pull request is currently `feature/rtodling/GEOSadas-5_29_3_4LDAS`, but it is easy to redirect the pull request to a different branch by clicking `Edit` in the top right corner of this page and then changing the target branch to Ricardo's latest working branch.

PR #121, opened 2021-11-01
Feature/rtodling/geo sadas 5 29 3

rtodling:
These include changes to allow running a full x0046a exp.


tclune (dismissed):
cmake changes ok.

mathomp4 (approved):
Approving!

rtodling:
Matt: hum ... that's odd that the branch is out of sync ... let me see what that's about

gmao-rreichle:
@rtodling, @mathomp4, 
Re. the changes related to the land-atm DAS (LADAS) in this PR, please merge the bug fix of PR #117 into Ricardo's branch **_before_** merging Ricardo's branch into develop.  It's just a few lines in `ldas_run.csh` and trivially 0-diff for the ADAS.

src/Applications/GEOSdas_App/fvsetup
```diff
@@ -3353,21 +3381,38 @@ sub set_newradbc {
...
-    #----------------------------------------------------
-    if (-e "$fvbin/LDASsa_assim_seq.x") {
+      $ans3 = query("LDAS HOME = $fvhome, full path?  ");
+      $ldashome = $ans3 ;
 
-        print "\n-------------\n";
-        print   "LDAS Analysis\n";
-        print   "-------------\n\n";
+      $ans4 = query("LDAS HOME for land ensemble = $fvhome/run/atmens, full path?  ");
+      $ldashome4ens = $ans4 ;
```
gmao-rreichle, line 3412: @rtodling: Here, is it really ok to suggest to the user that $fvhome could be the home directory for the LDAS?  If the user accepts $fvsetup as the LDAS HOME, they probably end up with a mess in $fvhome.  (I doubt that the ADAS and LDAS directories merge cleanly within the same directory tree.)
Rather than suggesting $fvhome, maybe we can construct a simple string that avoids the conflict, something concatenated like $fvhome+'_LDAS'   [in pseudo-code].

rtodling: I'm sorry what's the issue? All I see changing here a print statement. This is just a query for the user ... the user change/should answer whatever needed ... I just can't default to something that does exist!!

gmao-rreichle: I fully agree that we do not want to use a variable that did not exist.  But neither do we want to suggest to the user something that will potentially result in chaos.  A fix is in new PR #122.
cc: @saraqzhang

mathomp4 (approved):
The bits I can understand seem good. Trust in Ricardo. Approved!

PR #122, opened 2021-11-01
fixing suggestion of $fvhome as top-level LDAS dir

gmao-rreichle:
It doesn't make sense to suggest $fvhome as the top-level dir for the LDAS within the LADAS.  Replacing suggestion of $fvhome with $fvhome"_LDAS"


rtodling (approved):
Fine!

PR #123, opened 2021-11-02
incorrect perl statement from LDAS update

rtodling:
Two line bug fix - original change from Land folks were not correct perl statements.


mathomp4 (approved):
Approved

gmao-rreichle:
@rtodling: My mistake, sorry about the syntax error.  I'm confused, though, about including "$ldashome" in the first query, and "$ldasenshome" in the second query.

PR #128, opened 2021-11-12
update ana_gridcomp; trivial but needed fix

rtodling:
This fix a minor issue on the setup; and points to a patch to ana_GridComp that fixes Metop-A AMSUA turn off date.

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -46,7 +46,7 @@ FMS:
 GEOSana_GridComp:
   local: ./src/Components/@GEOSana_GridComp
   remote: ../GEOSana_GridComp.git
-  tag: v1.4.5.1
+  tag: v1.5.0
   develop: develop
 
 GEOSgcm_GridComp:
--- a/components.yaml
+++ b/components.yaml
@@ -46,7 +46,7 @@ FMS:
 GEOSana_GridComp:
   local: ./src/Components/@GEOSana_GridComp
   remote: ../GEOSana_GridComp.git
-  tag: v1.4.5
+  tag: v1.4.5.1
   develop: develop
 
 GEOSgcm_GridComp:


components.yaml
```diff
@@ -46,7 +46,7 @@ FMS:
 GEOSana_GridComp:
   local: ./src/Components/@GEOSana_GridComp
   remote: ../GEOSana_GridComp.git
-  tag: v1.4.5
+  tag: v1.4.5.1
```
gmao-rreichle, line 49: Maybe a silly question, but I thought the GEOS-ESM policy or guidance (from the SI team) was to have only three digits in releases/tags?   That is, I would have expected this to be v1.4.6 instead of v1.4.5.1
cc: @mathomp4

rtodling: Well, what's a patch?

gmao-rreichle: A patch is in the third digit.  See presentation #6 here (on GMAO Intranet):
https://gmao.gsfc.nasa.gov/intranet/presentations/git_transition/
But the guidance might have changed since then.
cc: @tclune, @mathomp4.

rtodling: Ok - I changed the tag to v1.5.0 since my earlier tag v1.4.5 where I had incremented the last digit from 4 to 5, but should have instead incremented the second digit.

tclune: I think we can be somewhat flexible about these distinctions - but it would be nice to have clear definitions for each repo.     My own thinking for MAPL has evolved somewhat recently.    We had been using "patch" for small updates and using "minor revision" for significant new features and/or to sweep up a bunch of small things.    MAPL is now using patch exclusively for bug fixes.     Then, for any given major+minor release, users should first be sure to have the latest patch if they encounter an issue.   Major release is for non backwards compatible changes - and yes there is  a MAPL 3 on the horizon.  (Doing lots of work to be sure GCM and ADAS work with that before we release.)

I'm also now leaning towards using a 4th digit for special circumstances to aid with ADAS integration.  (Just a side comment to @rtodling - this is what I alluded to in an email a couple of weeks back.)    The idea being that ADAS is esp. sensitive to changes.      If ADAS finds problems in the GCM that require tweaks, those would increment the 4th digit and the changes would be also be immediately merged into the development branches.  Works much like a hotfix, except it would not be off of the latest version.

tclune: I should add, some of the resistance to the approach outlined above is that it results in a much larger number of "minor" releases.   Of course, there is no shortage of positive integers, so this is mostly a mindset issue, but it does feel different when rolling out v3.217.5 of something.

tclune: I should also say that the [SemVer](https://semver.org) "standard" is X.Y.Z where X is major, Y is minor and Z is patch.   X defines the public API (backward compatibility constraint), Y is fo new features and other improvements.   Z is for bug fixes.  I had misinterpreted earlier readings  that allowed other minor things to go into patches.  To save you the trouble of finding the relevant bits I've pasted them here:

**Patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards compatible bug fixes are introduced. A bug fix is defined as an internal change that fixes incorrect behavior.**

**Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.**

**Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are introduced to the public API. It MAY also include minor and patch level changes. Patch and minor versions MUST be reset to 0 when major version is incremented._**

rtodling: The reason I thought of a 4th digit in this case is that the change was so small - really a small patch.  I think the presentation Matt pointed me to has it so I could have issued a beta version, but that does not sounds proper in this case since it's being released for OPS/FP. 

According to the definitions @tclune  just pasted here I should have just incremented the 3rd digit. Since I had already broken the rules above by not changing the middle digit in my previous tag, I issued a new that where now the middle digit reflects the changes.

tclune: @rtodling Thank you for being flexible as I learn and try to bring some uniformity to our processes.

rtodling:
Is this good now?

gmao-rreichle:
Certainly fine with me, but I'm not the one approving the PR.  I heard Matt is out of the office, but not sure.

mathomp4 (approved):

tclune:
@gmao-rreichle Note that Arlindo is developing a more nuanced SemVer description that attempts to address science changes.  Traditional SemVer is highly focused on interfaces, and creates significant ambiguity for how to uptick changes in science that leave interfaces unaffected.    Does not contradict what you've done here - just thought you might want to know.   Hopefully we'll be sharing this improved description soon.

PR #133, opened 2021-11-17
Add checking for prior calculation of virtual temperture before running 'prevents'

gmao-msienkie:
The new program check_virtmp.x will check the SFCSHP data in the input prepbufr file to determine if the virtual temperature correction has already been applied, and print out 'VIRTUAL' or 'DRY' to indicate the file state.  The 'gmao_prepqc' script has been modified to dynamically create the input parameter file for 'prevents.x' based on the determination made by check_virtmp.x - i.e. the virtual temperature correction will be set to FALSE if the correction has already been applied to the input files.  

This will correct a problem with the MERRA2 prepbufr files that started at the end of December 2020.  Previously the configuration in their processing had not done the virtual temperature calculation for SFCSHP and ADPSFC prior to sending us the files.  The double calculation of virtual temperature caused the input marine temperatures for MERRA2 to be warmer than the operational marine temperatures and may have contributed to the excessive warm tempertures in the western Atlantic that were discussed at the monitoring meeting.

This change will be zero diff for the prepQC processing until 2020/12/29 00z which is when the input stream switched to virtual temperature from dry temperature for SFCSHP and ADPSFC.


tclune (dismissed):
CMake changes OK.

jing-guo (dismissed):
No objection, but not merging the PR.

gmao-msienkie:
Sorry - I realized when I was adding this code to the CVS repository for the MERRA2 and CERES runs that I had not included a 'usage' statement to be printed if a command-line argument was not specified for the input file.  I guess you'll have to review again.  Oh, well.

gmao-msienkie:
> No objection, but not merging the PR.

That's fine.  This change is not needed for regular forward processing and similar tests since we don't use prepQC for those runs, we just ingest the GDAS prepbufr.  
There is a pending request for a prepQC change to add the NRL ACQC that will be needed for R21C and possibly GEOS-IT (the segments that use historical data).  Both that branch and this one modify the 'gmao_prepqc' script so we'll have to take some care to make sure both changes end up getting in.

rtodling:
Meta, I updated your newer PR (for the aircraft stuff) before accepting this ... so there was conflict in the script - which I think I resolved; but this PR now needs your ok ... can you please look at it? Thanks

gmao-msienkie:
It looks like the merge with develop (with the ACQC change) is good.  I'm setting one additional switch in the parm file to have the program not recalculate Q.

rtodling (approved):
fine w/ me

Issue #119, opened 2021-10-26
Check consistency between job input files and fvsetup


gmao-jstassi:
In CVS, the runjob script checked for tag consistency with the input file before executing. This feature was lost when the code was moved to the git repository. It will be replaced by a consistency check between the input file and the fvsetup script. This is actually a better check since several DAS tags can share the same version of fvsetup, so it doesn't matter if the tag changes if fvsetup hasn't changed.

Issue #83, opened 2021-04-13
AOD parallel jobs


gmao-jstassi:
The AOD job processing logic in the GEOSdas.csm file needs to be cleaned up.

In the 3DVar case, AOD can be run in parallel with GSI if enough CPUs are available for both. Otherwise, AOD should be submitted as a separate job, and GCM processing should be held up until AOD completes. This logic is pretty much already in the code.

In the ensemble case, AOD needs to be run prior to GSI processing. There is no need to run it as a separate job. It just needs to be run and completed prior to starting the GSI. It cannot be run in parallel with the GSI.

PR #137, opened 2021-11-30
Feature/jstassi/#83/#119/aod parallel and runjob consistency check

gmao-jstassi:
Issues
#83 Logic to run AOD in parallel and wait rather than running as a separate job
#119 Updates for runjob to check consistency between input file and fvsetup


tclune (approved):
cmake changes ok.   The name "CVSTAG" seems a bit dated now though.

gmao-jstassi:
The "CVSTAG" variable was already in the file and was not added with the updates; though there was a line format modification (but no code change) which would cause the name to show up in a list of diffs.

rtodling (approved):
These seem fine w/ me.

PR #138, opened 2021-12-02
Feature/msienkie/#33 implement nrl acqc

gmao-msienkie:
The NRL ACQC will be needed for the GEOS-IT processing, to supply aircraft profile files for assimilation with the variational aircraft temperature bias correction.  (The NCEP profile files are not available prior to 2016.)

This hasn't been as much an issue with the prior DAS releases because people were not running the variational aircraft bias correction for dates earlier than 2016 and were primarily running with GDAS input data.   For GEOS-IT, R21c and other retrospective analyses this will need to be implemented.


gmao-msienkie:
The assimilation will be zero-diff for running with the GDAS data (without the prepQC) but will produce different (and likely improved) results when running prepQC.

tclune (approved):
CMake changes ok

rtodling (approved):
This is fine with me. I will pull this into what is forming 5.29.3_p3; but will have to pull this into the branch that presently supports 5.27.1 which is presently being used for GEOS-IT; I strongly recommend the branch this sits under not to be removed.

gmao-msienkie:
I have a copy of GEOSadas-5_27_1p3 with this branch in my /discover/nobackup/projects/gmao/obsdev/msienkie/R21C_git_build/GEOSadas that I put together back in the summer.  
I am hoping that people will not want to use the old acqc/acarsqc because I don't want to have to support it since it is long gone from NCEP.  Meanwhile, if people want to "run like MERRA2" they can also pull the saved d5124_m2_xxx.prepbufr files from the archive and use exactly the input obs that were used for MERRA2, and not have to run prepQC at all.

PR #139, opened 2021-12-06
Feature/msienkie/acft bc cleanup tails

gmao-msienkie:
Modifications made to 'GEOSdas.csm' and 'fvssi' scripts to add the 'cleanup_tail=.true.' flag to the options for the two aircraft varBC methods.  
The cleanup_tail process removes any entries from the aircraft temperature bias correction file that have not been updated for more than a year.  This helps limit the size of the aircraft coefficient file.

When a particular tail number has no reports the bias correction coefficients and their background error are kept the same until that particular tail number resumes reports again - the background error is not inflated.  Thus, this change will not be zero diff in the case where an aircraft reappears.  If a tail has been missing for more than a year it is probably beneficial to discard the bias correction since the characteristics of the aircraft equipment may have changed or the equipment may have been replaced during the downtime.

### DIFF
--- a/src/Applications/GEOSdas_App/GEOSdas.csm
+++ b/src/Applications/GEOSdas_App/GEOSdas.csm
@@ -913,11 +913,11 @@ exit 1
 	  echo 'Setting aircraft_t_bc_ext to true, using external bias correction'
 	  breaksw
      case 2:
-          echo "s/>>>AIRCFT_BIAS<<</aircraft_t_bc=.true.,/g"  >> sed_file
+          echo "s/>>>AIRCFT_BIAS<<</aircraft_t_bc=.true.,cleanup_tails=.true./g"  >> sed_file
 	  echo 'Setting aircraft_t_bc to true, using VV.VV^2 bias correction'
 	  breaksw
      case 3:
-          echo "s/>>>AIRCFT_BIAS<<</aircraft_t_bc_pof=.true.,/g"  >> sed_file
+          echo "s/>>>AIRCFT_BIAS<<</aircraft_t_bc_pof=.true.,cleanup_tails=.true./g"  >> sed_file
 	  echo 'Setting aircraft_t_bc_pof to true, using POF bias correction'
 	  breaksw
      default:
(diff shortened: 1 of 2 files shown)


gmao-msienkie:
The cleanup_tail action is now optional, controlled via an environment variable.  
setenv CLEANUP_TAIL 1  ->  cleanup_tail=.true.        
if the environment variable is not set, the default is not to set cleanup_tail, and the GSI by default sets that as false.

rtodling:
Meta,

Do we really need to make this optional? That is to have an env variable controlling this?

rtodling (commented):
... I think  I understand what is controlled by this env var; but should we make the default env var 1 instead of 0 ... ideally we don't have to tell OPS to add another env var. And 0 would only be needed in cases we are trying the reproduce  a previous run, no?

rtodling (commented):
Actually, I am looking again and I am realizing that the way you changed this, when the env var is set to zero the gsi.rc after being edited will have a line:

">>>AIRCFT_BIAS<<<"

like this in the namelist and the fortran code is going to choke! We need to define the if differently ...

gmao-msienkie:
I tested this in a run.  When I had the enviroment variable set to 1 the CLEANUP_TAIL was set to TRUE.  When I commented out the definition, the first time around it had CLEANUP_TAIL as TRUE because I think the job inherited the environment variable or something.  When I independently submitted a segment it set the undefined CLEANUP_TAIL as FALSE.  I am trying to run a segment where I have the environment variable set to 0 now so I can check what it does. 

My code looks like this - it continues to set the >>>AIRCFT_BIAS<<< as needed and then appends the value of cleanup_tail to the substitution string if the environment variable is set to true (1).

     case 2:
          set cftstring = "aircraft_t_bc=.true.,"
          if ( $CLEANUP_TAIL ) set cftstring = "$cftstring cleanup_tail=.true.,"
          echo "s/>>>AIRCFT_BIAS<<</$cftstring/g"  >> sed_file
          echo 'Setting aircraft_t_bc to true, using VV.VV^2 bias correction'
          breaksw

rtodling:
Sorry - I missed it ... all fine!

rtodling (approved):
ok - sorry - all good

PR #140, opened 2021-12-13
Feature/rtodling/berror flip fix

rtodling:
Minor fixes in various scripts; addition of input file for Summer of x0046a; and major fix in tranfs - none of its programs worked in the previous version.


mathomp4:
Oops. Looks like Tom pulled in my CMake fix (#141) and so this is now out-of-date. Didn't realize it was something he could do. @rtodling if you can update and make sure all is well?

(Note that #141 can't affect anyone as you have to work *really hard* to get to the other side of the code path it fixes.)

PR #147, opened 2021-12-16
        (1) add ldas anal increments to fcst script

saraqzhang:
(1) add ldas analysis increments to g5fcst script that is generated from fvsetup 
        (2) fix syntax errors in the current develop fvsetup

### DIFF
--- a/src/Applications/GEOSdas_App/fvsetup
+++ b/src/Applications/GEOSdas_App/fvsetup
@@ -3796,9 +3796,11 @@ sub get_obsys {
   elsif ($ans == 3) {  # MERRA-2
       $obsysrc = "$fvetc/obsys-${loc}-merra2.rc";
       $rflags .= " -stem merra2";
+  }
   elsif ($ans == 4) {  # GEOS-IT
       $obsysrc = "$fvetc/obsys-${loc}-geosit.rc";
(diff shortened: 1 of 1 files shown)


gmao-rreichle:
@rtodling , @gmao-jstassi : Here's a quick summary of this PR:

1. The PR fixes a bug in fvsetup that was introduced recently in an unrelated PR (nothing to do with the land).  You may already have fixed this bug separately.  

2. The PR further makes a trivially 0-diff change that applies increments from the LDAS to forecasts during the first few hours, which are just a re-run of the Corrector segment (similar to applying IAUs during the initial period).  This change applies only when the ADAS is coupled to the LDAS (hence trivially 0-diff).  According to @saraqzhang, this change needs to be made in fvsetup, which is not ideal because fvsetup is already bulky.  Let us know if you can think of a different way to implement this.  

3. Besides the verification in the coupled ADAS-LDAS system, we did not test the PR because it seems to be trivially 0-diff.  Let us know if you need us to do additional testing. 

There's no rush in merging this PR, but it would be good not to keep it in limbo for very long to avoid having to maintain it as "develop" evolves.   Let us know if you have any questions, thanks, - Rolf

rtodling (approved):
ok

PR #150, opened 2021-12-23
update GSI for GEOS-IT config; minor script adjustments merging 5.27.…

rtodling:
…1-IT with current


rtodling:
This is zero-diff for FP-like settings; but non-zero diff for GEOS-IT

mathomp4 (approved):

Issue #161, opened 2022-03-01
Updates to 5_29_4 tag


gmao-jstassi:
(modified) monthly_means.pl: Modified to use get_htype() routine from Manipulate_time.pm
(modified) monthly_tarandclean.j.tmpl:  Modified to get fileTemplate using a label which is replaced with value 
(new file) testsuites/update_input_IDs.pl:  This is a new file that will update the fvsetupID and codeID values in the testsuites *.input files

PR #162, opened 2022-03-02
Modified in GEOSdas_App: monthly_means.pl and monthly_tarandclean.j.tmpl

gmao-jstassi:
issue #161
The monthly_means.pl script was modified to use get_htype() routine from Manipulate_time.pm.
The monthly_tarandclean.j.tmpl was modified to get fileTemplate using a label which is replaced with value.
A new file in GEOSdas_App/testsuites, update_input_IDs.pl, is a utility which can be used to update the testsuites/*.input files so that they will pass the consistency test in the runjob script.


rtodling (approved):
I will assume this works!

Issue #166, opened 2022-03-09
Problem with traj_lcv_rst entry in the DAS run directory resource files


gmao-jstassi:
During DAS job setup, the traj_lcv_rst collection is not being handled correctly, which affects how it is added to the following files: 
- monthly.arc
- monthly.keep.arc
- monthly.rc
- mstorage.arc
- silo.arc

The solution is a one-line fix in the edhist.pl script.

PR #167, opened 2022-03-09
bug fix to handle traj_lcv_rst correctly

gmao-jstassi:
#166
This update will correct how the traj_lcv_rst collection is handled in the experiment run directory resource files: monthly.arc, monthly.keep.arc, monthly.rc, mstorage.arc, and silo.arc


rtodling (approved):
Thank you Joe.

PR #169, opened 2022-03-18
Add code updates to handle GEOS-R21C configuration

elakkraoui:


tclune (approved):
cmake changes ok

rtodling (commented):
I am looking at the PR in GEOSadas … and it still have files that don’t need to be included!

_setup_4geos.pl
_setup_4r21c.pl

I file fvsetup: I don't understand the change in line 1219 ... that's not to be commented. The original line is correct .. .when the default is set the other variables don't need to be set ... they are zero by default!

What are the files

src/Appl..../gmao/etc/R21C/test.txt
and test3.txt
?? I don't think these should be committed or in the PR

PR #174, opened 2022-04-18
Feature/msienkie/prep qc modifications

gmao-msienkie:
Requesting changes for PAQC to be merged into 'develop' and future tags.

The impetus for this request is to add the NCEP changes in CQCVAD need to process the new 'Level 2' data - increases in array sizes.   Should be zero diff prior to the new data start (3/30/2022).

There is a small bug fix in fix_ascat.f90 to correctly implement a quality control check.   I believe this program is only needed when running with gdas_prepbufr files during the period before 2011/05/09 12z when the ASCATW data was given a purge flag.

Modified the formatting in a few routines - gsidiag_bin2txt.f90, scanbuf2.f

Add some new routines used in MERRA2  flag_NPbuoy.f, twindow.f (twindow.rc, twindow_m2.rc)  Also added utility program scanbuf0_accum.f (like scanbuf0 but only prints cumulative counts at the end).  

Merged in NCEP changes to bring code up to date with latest NCEP (obsproc_prep.v5.5.0) - prevents.f, cqcbufr.f cqcvad.f profcqc.f

Modified ancient copy of 'read_prepbufr' code in radcor directory to work with current prepbufr format files.

Modified oiqcbufr.f to use 10-digit dates (yyyymmddhh) with the BUFR library .  This avoids future problems with the shifting time window for 8-digit dates.


gmao-msienkie:
Sorry about including all the miscellaneous files - if necessary I can pull some of these off onto separate requests.  After the CQCVAD problem I thought it would be a good idea to catch up with changes that I had been accumulating, to bring us up to date with NCEP

tclune (dismissed):
cmake changes ok

gmao-msienkie:
Technically the changes are non-zero diff.  It is extremely unlikely for the modification in fix_ascat to trigger a change in the output and more than likely that the data formerly unflagged would be used in the assimilation.
The changes in cqcvad.f to increase the array size are only non-zero after the start of the new data at the end of March 2022.

gmao-msienkie:
I ran some tests with this code and compared with results from running with code from a 5_29_4 tag.  It looks like the cqcvad.f already had changes to increase array sizes versus the code in MERRA2.  The BUFR files resulting from running gmao_prepqc were zero diff.   It would still be good to bring in the various prepqc changes in this tag so we can be prepared for possible future changes.

rtodling:
Meta, I am bring this in, ok?

gmao-msienkie:
I modified CQCVAD to increase the maximum number of reports, and I tested the code to see there are zero diffs for files with less than 50000 levels and that files with more than that number of levels are handled correctly by the new code.
I merged the current develop branch and pushed my changes to GitHub.

rtodling (approved):
Thank you Meta.

PR #175, opened 2022-05-03
Update the readme file with a note about the 'generic' configuration

gmao-msienkie:
Updated the readme_init_coeffs file with a note about how it uses the 'gsi.rc.tmpl' to get 'dtype', 'dplat' and 'dsis' for each satellite.


gmao-msienkie:
Whoops - now adding a code change as well.   Results should be zero diff - except for geoimr which is not used in the standard build.

rtodling (approved):
Thanks Meta.

PR #178, opened 2022-05-10
The script was modified to add an hourtype value for the traj_lcv_rst…

gmao-jstassi:
… entry.


rtodling:
Joe, 

This seems quite a large change as compared to the simple explanation given above: 
>> The script was modified to add an hourtype value for the traj_lcv_rst… <<

I assume these are zero diff changes - I see no reason why they'd not be.  I am assigning a zero-diff label to this ... I know you labeled bug fix ... but you always need to indicate zero or non-zero diff.

Thx

rtodling (approved):
I assume you tested this so I will approve - and test it also. 

Thanks.

PR #184, opened 2022-06-03
Modify prepqc script to check for pre-qc data and skip if no data is present

gmao-msienkie:
Originally the prepqc script determined the templates for the 'pre-qc' files to be combined and passed through 'prepqc' by checking the output of the 'acquire_obsys' command.  A few years ago the DAS was modified to obtain the templates for the 'pre-qc' files by checking the OBSCLASS and obtaining the templates from the obsys.rc.  This was necessary because the 'acquire_obsys' was being run several times and the obsys.acq output from e.g. the acquire of aerosol data was overwriting the the acquire of the GSI observations. 
The problem with this is the 'obsys.acq' had previously only had filename templates for data that was actually acquired for the segment while the newer method obtained templates for all of the files with pre-qc OBSCLASS entries even if the data was not being obtained.
This script adds a check that each pre-qc file exists before the filename is added to the input to the 'combfrd.x' program that combines the prepbufr file.  If there are no matching pre-qc files present, the rest of the prepqc related processing is skipped.
Among other things, this allows a DAS run to transition between MERRA2/GEOSIT/R21c reanalysis data streams to the real-time GDAS stream (which does not need and should not have 'prepqc' processing performed on it) without having to stop and reconfigure.  The obsys.rc file can be configured to use the 'pre-qc' files up to a certain date and the real-time files after that date.
The GEOSdas.csm modifications have been tested.  There are similar modifications in 'prepqc_daemon.pl' that have not been tested since it doesn't look like that script is being used in current runs.

I have also made a small change that will hopefully remove the dependency on having the unneeded ana_satbang_rst file in the restart files.  A test has been set up running with the old VarBC to verify that the ana_satbang_rst file will be handled correctly for that case.

### DIFF
--- a/src/Applications/GEOSdas_App/GEOSdas.csm
+++ b/src/Applications/GEOSdas_App/GEOSdas.csm
@@ -1969,9 +1969,11 @@ exit 1
         endif
      endif
 
-     set ars_list = ( satbias satbang )
+     set ars_list = ( satbias )
      if ($NEWRADBC || $ANGLEBC) then
        set ars_list = ( $ars_list satbiaspc )
+     else if ( ! $ANGLEBC ) then
+       set ars_list = ( $ars_list satbang )
      endif
      if ($ACFTBIAS) then
        set ars_list = ( $ars_list acftbias )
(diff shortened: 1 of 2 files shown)


gmao-msienkie:
The test with unified radianceVarBC showed that the satbang rs was no longer needed.  The test with the old bias correction (and $ANGLBC=0) required and acquired the satbang rs as needed

rtodling (approved):
I get the sense that your changes in GEOSadas associated w/ satbang need similar changes elsewhere ... I take this in and see what I get. Thanks

PR #186, opened 2022-06-03
Bugfixes/jstassi/#185/fvsetup

gmao-jstassi:
* setenv ARCHIVE in the fstats_run script, so that the script can be run in an environment where ARCHIVE is not defined correctly for the experiment
* change "tail +2" syntax to "tail -n +2"

### DIFF
--- a/src/Applications/GEOSdas_App/fvsetup
+++ b/src/Applications/GEOSdas_App/fvsetup
@@ -8331,9 +8331,9 @@ print  SCRIPT <<"EOF";
 
           cd \$FVWORK
           if (`uname -n` =~ borg*) then
-             cat \$PBS_NODEFILE | head -1 > \$FVWORK/OIQC_list
-             cat \$PBS_NODEFILE | tail +2 > \$FVWORK/GCM_list
-             cat \$PBS_NODEFILE           > \$FVWORK/ANA_list
+             cat \$PBS_NODEFILE | head -1    > \$FVWORK/OIQC_list
+             cat \$PBS_NODEFILE | tail -n +2 > \$FVWORK/GCM_list
+             cat \$PBS_NODEFILE              > \$FVWORK/ANA_list
              subset_nodefile.pl \$NCPUS_IDF -f \$PBS_NODEFILE -o \$FVWORK/IDF_list
              subset_nodefile.pl \$NCPUS_IAU -f \$PBS_NODEFILE -o \$FVWORK/IAU_list
           endif
@@ -8610,6 +8610,7 @@ EOF
            endif
            echo2 ""
            echo2 "setenv FVHOME \$FVHOME"
+           echo2 "setenv ARCHIVE \$FVARCH"
            echo2 "set Bin = \$Bin"
            echo2 "set output = \$lname2"
            echo2 ""
--- a/src/Applications/GEOSdas_App/testsuites/C180RPY.input
+++ b/src/Applications/GEOSdas_App/testsuites/C180RPY.input
@@ -2,9 +2,9 @@
 # C180RPY.input
 #------------
 
-codeID: b3a880f
+codeID: 523f29e
 description: C180RPY__GEOSadas-5_29_3__agrid_C180__ogrid_C
-fvsetupID: f7d8d041c9
+fvsetupID: f7aaa973c7
 
 ---ENDHEADERS---
 
--- a/src/Applications/GEOSdas_App/testsuites/C180T14RPY.input
+++ b/src/Applications/GEOSdas_App/testsuites/C180T14RPY.input
@@ -2,9 +2,9 @@
 # C180T14RPY.input
 #------------
 
-codeID: b3a880f
+codeID: 523f29e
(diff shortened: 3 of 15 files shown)


rtodling:
Hi Joe,

I see that you are wiring in the git ref git (tag) in the input files ... this is a bit of a nightmare. We need to find a way to centralize this somehow and not have to change each and every one of these files just because we made a new commit to GEOSadas fixture that has nothing to do w/ the input files. Can you think of something?  I will take these as they are here for now, but as soon as I put another change to develop this will fall out of sync.

mathomp4:
@rtodling I can say it is possible to fill these in at CMake time. With the right code, CMake can figure out what the git commit of its own source code is and then fill that in. For example, if I add to the GEOSgcm `CMakeLists.txt`:
```cmake
find_package(GitInfo)
GIT_WC_INFO(${CMAKE_CURRENT_SOURCE_DIR} GEOSgcm)
```
this will fill in a bunch of variables as defined in the [`FindGitInfo.cmake`](https://github.com/GEOS-ESM/ESMA_cmake/blob/main/external_libraries/FindGitInfo.cmake) macro we carry in ESMA_cmake. If I then print them out with:
```cmake
include(CMakePrintHelpers)
cmake_print_variables(GEOSgcm_WC_REVISION_HASH)
cmake_print_variables(GEOSgcm_WC_REVISION)
cmake_print_variables(GEOSgcm_WC_REVISION_NAME)
cmake_print_variables(GEOSgcm_WC_URL)
cmake_print_variables(GEOSgcm_WC_ROOT)
cmake_print_variables(GEOSgcm_WC_LAST_CHANGED_DATE)
cmake_print_variables(GEOSgcm_WC_LATEST_TAG)
cmake_print_variables(GEOSgcm_WC_LATEST_TAG_LONG)
```
then at CMake time we see:
```
-- GEOSgcm_WC_REVISION_HASH="9e87124"
-- GEOSgcm_WC_REVISION="9e87124"
-- GEOSgcm_WC_REVISION_NAME="9e87124 main"
-- GEOSgcm_WC_URL="git@github.com:GEOS-ESM/GEOSgcm.git"
-- GEOSgcm_WC_ROOT="git@github.com:GEOS-ESM/GEOSgcm.git"
-- GEOSgcm_WC_LAST_CHANGED_DATE=""2022-07-20 15:58:27 -0400""
-- GEOSgcm_WC_LATEST_TAG="v10.22.3"
-- GEOSgcm_WC_LATEST_TAG_LONG="v10.22.3-13-g9e87124"
```

You could use this in combination with [`configure_file`](https://cmake.org/cmake/help/latest/command/configure_file.html) to insert into scripts. I did try something like this in the [testsuites CMake](https://github.com/GEOS-ESM/GEOSadas/blob/develop/src/Applications/GEOSdas_App/testsuites/CMakeLists.txt) when I converted from CVS to Git, though I might have gotten some things wrong.

gmao-jstassi:
Ricardo,

It is not as bad as you think.

The reason I updated the ID values in the input files for this PR is because I modified the fvsetup script.

I did not update the input files individually. There is a utility script in the GEOSdas_App/testsuites directory called update_input_IDs.pl which updates all the input files at once. It even finds the needed values for you, so it is very easy. You should try it.

Here is an explanation of the ID values in the input files:

The input files have two IDs, codeID and fvsetupID. The codeID is set to a GIT_TAG_OR_REV variable that Matt created to replace the CVS TAG info when we converted to git. Currently, codeID is not used for anything, and I have considered removing this variable from the input files.

The fvsetupID is the git ref ID for the fvsetup script. This value is used for comparison between the runjob script and the input file. The idea is that if fvsetup has changed, then the prompts may not match those in the input files. All we really care about is updates to fvsetup. We don't care about any other updates in the system.

So the only time the ID values in the input files need to be updated is when fvsetup is modified, and  even then it is not critical, because the runjob script will give the user the option to continue, even if the IDs don't match.

What we could do if only update the ID values when there is a new tag or release. Then we won't be doing any more work than what we did when we were using CVS, and it will actually be easier because of the update_input_IDs.pl utility.

If you still have questions about this, then let me know.

Joe

________________________________
From: rtodling ***@***.***>
Sent: Thursday, July 21, 2022 12:43 PM
To: GEOS-ESM/GEOSadas ***@***.***>
Cc: Stassi, Joe (GSFC-610.1)[SCIENCE SYSTEMS AND APPLICATIONS INC] ***@***.***>; Author ***@***.***>
Subject: [EXTERNAL] Re: [GEOS-ESM/GEOSadas] Bugfixes/jstassi/#185/fvsetup (PR #186)


Hi Joe,

I see that you are wiring in the git ref git (tag) in the input files ... this is a bit of a nightmare. We need to find a way to centralize this somehow and not have to change each and every one of these files just because we made a new commit to GEOSadas fixture that has nothing to do w/ the input files. Can you think of something? I will take these as they are here for now, but as soon as I put another change to develop this will fall out of sync.

rtodling (approved):

PR #189, opened 2022-06-23
Update various reanalysis-related resource files for GEOSIT and R21C

elakkraoui:


rtodling (approved):
There is some redundancy between this ticket and another from Meta (#184).

PR #192, opened 2022-07-25
reconfig RO top

rtodling:
This reconfigures the RO top to 55 km; and sets up a neutral option for commercial RO observation error scaling (viz. x0046d).

This is a non-zero diff change. Original use of RO to 60 km is in large disagreement and discrepancy w/ what experts in RO think of the data above 55km.


mathomp4 (approved):
Approve!

rtodling:
Thanks Matt

PR #196, opened 2022-08-23
updates to fix aerosol assimilation of AVHRR data for GEOSIT/R21C

elakkraoui:
Adds update to enable aerosol assimilation from historical AVHRR data for GEOSIT.


rtodling (approved):
All good - thanks.

PR #199, opened 2022-08-23
Feature/jstassi/#198/local acquire gaas obs

gmao-jstassi:
Updates to handle local acquire of gaas_obs data using entries in the obsys-gaas.rc file.


rtodling (approved):
I will give this a try ...

PR #207, opened 2022-09-06
Second round of fixes to GEOSIT resource files

elakkraoui:


rtodling (approved):
These seem fine ... but proof will be when I run a test on it

rtodling:
Matt, can you tell what's failed here?

PR #216, opened 2022-09-30
vb:update ana_aod.F code to handle both AOD and LAOD obs at the same …

vbuchard:
- Updated ana_aod.F code to be able to handle AOD (KT=45) and Log of AOD (KT=43) observations at the same time.
- The code should work if only AOD is available or
                                         if only LAOD is available or
                                         if both of them are available at the same time.

### DIFF
--- a/src/Applications/GAAS_App/ana_aod.F
+++ b/src/Applications/GAAS_App/ana_aod.F
@@ -67,9 +67,10 @@
 !     ---------------
       integer :: im, jm, km, nch
       integer :: im_so, jm_so    ! dimension of superob space
-      integer :: rc, nobs, nobs_good, nODS, i
+      integer :: rc, nobs, nobs_good, nODS, i, naod, nlaod
       integer :: nymd=0, nhms=0, nymd_b, nhms_b
       integer :: myKTobs
+      integer :: kt_list(2)
 
       real*8  :: alpha         ! bias update coefficient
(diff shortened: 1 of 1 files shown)


rtodling:
Virginie, you'll need to add a zero-diff or non-zero-diff label, please.

rtodling:
Also, in order to get zero-diff w/ the code that was there before you need to convert the data in exactly the same way as the observer is converting the obs:

 ods%data%obs = max(eps,exp(ods%data%obs)-eps)

you should not do:

 ods%data%obs = max(eps,exp(ods%data%obs)-0.01)

eps is a resource file parameter.

Also, how did you test? Out of the three possible cases now, did you make sure that the two previous cases give you zero diff in the increments?

vbuchard:
I tried that first by using eps but eps is not defined in ana_aod.F, so I had differences in the observations values in my test. By putting the value of eps directly in the code as being 0.01, I had 0 diff. eps is defined in the observer code m_obs.F90. 

For the tests, I run ana_aod.F, followed by PSAS offline. I did not look at the increments directly since I did not touch this part of the code. I looked at how many observations were going through to the qc and if the observation values in the post-processing obs files were similar to Amal's experiment files. To look at the increment values, I need to be sure that the same background files are used.

rtodling (approved):
This seems fine w/ me - and I know the tests worked ... taking it on.

PR #226, opened 2022-10-17
Fix for RO FSOI diag.bin issue

rtodling:
Update other repos - no change to fixtures:

GEOSana - fix for RO diag.bin error that made FSOI got bad
GMAO_Shared - update restricted RO database (consistent w/ x0046d - summer)
GEOSgcm_App - revise regrid-method for few FP streams

zero diff as long as FSOI is not running; otherwise it fixes an error there.


mathomp4 (approved):

rtodling:
Thank you.

PR #234, opened 2022-11-30
Fix mp_stats.x

### LINKED ISSUE #233: mp_stats.x broken
User @gmao-emcgrath was testing some particular GEOSadas workflow flavor that apparently exercises mp_stats.x, looks like this is broken if the input and output resolutions do not match (which means a regrid is done underneath). The issue is that the grid is not created via the MAPL grid manager.

In addition while testing this I found a flat out bug in the gbl2lcl_ routine, it was using a variable jm that should be jn. I will fix too.

bena-nasa:
Fixes #233


rtodling (approved):
Ok - let's see if this solves the issue I have after merging w/ v10.23.0 ... and corresponding MAPL

PR #235, opened 2022-12-01
Move to use GitHub Action for Label Enforcement

mathomp4:
Due to changes at Heroku, we need to transition to use a GitHub Action for label enforcement. The current label enforcer no longer works, so PRs will be "stuck" until this is fixed.


rtodling (approved):
You know better ...

PR #237, opened 2022-12-06
Feature/vbuchard/add fraction of brown carbon

vbuchard:
At the time of deriving the fractional composition, if a brown carbon tracer is available, the code now adds the fraction of Brown Carbon to the fraction of Organic Carbon (foc). It then also recomputes the fraction of total carbon (fcc=foc+fbc).


rtodling (approved):
Ok - I had missed this ... good. Thank you.

Ricardo

rtodling:
Actually, I approved this but I am thinking that this might not be quite right!  I think that instead of 

self.fcc  = self.fbc + self.foc

your need:

self.fcc  = self.fbc + self.fcc

Since fcc includes both oc and bc ... the way you did it you seem to be bypassing the contribution from OCEXTTAY  !! ??

PR #244, opened 2023-02-13
Fix tag for GMAO_Shared

mathomp4:
I noticed a bug in `components.yaml` for GMAO_Shared.


rtodling (approved):
ough! my bad ... copy and paste gone wild

PR #245, opened 2023-02-13
Add CI Build to GEOSadas

mathomp4:
Just realized we don't have a CI GEOSadas build here. This was inspired by #244

We should do that and I think this will work...


mathomp4:
Seems to work! Huzzah!

tclune (dismissed):

rtodling (approved):
Ok

PR #253, opened 2023-06-06
GEOSadas should ignore Held-Suarez related files

mathomp4:
Recently in the GEOSgcm, @pchakraborty added support for the Held-Suarez model (used in GPU work) but it is *ignored* by the GEOSgcm by default by sparsing it out (see https://github.com/GEOS-ESM/GEOSgcm/pull/591). 

This PR brings the same sparsing here. This was uncovered in MAPL CI as the GEOSadas builds were all failing on some OpenACC calls because it was seeing the Held-Suarez code.

### DIFF
--- a/config/GEOSgcm_GridComp.sparse
+++ b/config/GEOSgcm_GridComp.sparse
@@ -1,3 +1,5 @@
 /*
+!/GEOSagcm_GridComp/GEOS_AgcmSimpleGridComp.F90
+!/GEOSagcm_GridComp/GEOShs_GridComp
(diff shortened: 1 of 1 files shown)


tclune (approved):

tclune:
@rtodling  I accidentally hit merge on this PR.   (It was mixed in with a queue of PR's for MAPL.)    I think it is harmless from your perspective, and it was urgently needed to restore the CI capabilities in MAPL.   Scream and we can roll it back.

PR #265, opened 2023-10-18
        updates for handling forcing outputs in ladas cycling

saraqzhang:
modified:   src/Applications/GEOSdas_App/GEOSdas.csm
	modified:   src/Applications/GEOSdas_App/fvpsas

### DIFF
--- a/src/Applications/GEOSdas_App/GEOSdas.csm
+++ b/src/Applications/GEOSdas_App/GEOSdas.csm
@@ -5865,13 +5865,6 @@ endif
               /bin/cp $EXPID.ana_radstat_rst.$rtag.tar $RSTHOLD/ &
             endif
 
-            # hold lfo files for the next segment of ldas coupling
-            #----------------------------------------------------------
-            #if ( $LDAS_ANA ) then
-               mkdir -p              $FVHOME/lana//holdforc
-               /bin/cp  $FVHOME/recycle/holdpredout/*2d_lfo*nc4  $FVHOME/lana/holdforc/
-            #endif
-
             # local copy to be moved by pesto, 2nd copy to be dealt w/ by recycle
             #--------------------------------------------------------------------
             if ( -e biasinp.$RSTSUFFIX ) then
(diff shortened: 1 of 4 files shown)


src/Applications/GEOSdas_App/GEOSdas.csm
```diff
@@ -5866,10 +5867,10 @@ endif
 
             # hold lfo files for the next segment of ldas coupling
             #----------------------------------------------------------
-            if ( $LDAS_ANA ) then
-               mkdir -p              $FVHOME/recycle/holdforc
-               /bin/cp  *2d_lfo*nc4  $FVHOME/recycle/holdforc/
-            endif
+            #if ( $LDAS_ANA ) then
```
gmao-rreichle: Is the if condition here intentionally commented out?   Based on a quick look it seems that the if condition should be there, but I might be wrong.

saraqzhang: @gmao-rreichle ,   that block was leftover from before, I forgot to delete it in the update. I will do it now the push the change.  thanks for spotting it.

gmao-rreichle (dismissed):

gmao-rreichle:
Looks ok to me, but I'm not the judge here.
I changed the label to "0 diff trivial" because the edits are all within `if ( $LDAS_ANA )` blocks.
@rtodling, from what I understand, this fix is needed because recent ADAS changes broke the weakly-coupled land assimilation.

mathomp4:
The force push above was:
```
gh repo clone GEOS-ESM/GEOSadas GEOSadas-Reset
cd GEOSadas-Reset
git checkout feature/saraqzhang/ladasupdate
git reset HEAD~5
git add .
git commit -m "fixed typo for zeit_c?.x MovePredictorOutput (fvpsas)"
git push --force origin feature/saraqzhang/ladasupdate
```
The add was for the "fix misspelling" commit that was previously there.

saraqzhang:
@gmao-rreichle @rtodling     
I updated the prompts and defaults concerning LDAS coupling in fvsetup and added the sample input file for ladas setup into testsuites/

saraqzhang:
the changes here are consolidated into PR #275.

PR #269, opened 2024-02-22
Feature/rtodling/atmos 11 5 1 for das

rtodling:
This is a prep for the release of a 5.40 tag which corresponds to mainly an update to the model version v11.5.1.

In this, most of the last changes in the analysis are kept untouched, except that CRTM-2.4.0 is backed away.


mathomp4:
Ah. The CI in this branch is now out of date. I can fix that up, @rtodling. Let me know if you are fine with that (mainly because you'll need to do a `git pull` on your end before pushing any more changes).

PR #278, opened 2024-06-05
Enable handling of cubed-grid background

rtodling:
This is still not working due to some issue in MAPL_cfio ...


rtodling:
No this goes a little further now that I put GFIO back (in place of CFIO that fails) and I put a hack to fix the dims accordingly since GFIO doesn't get jm world correctly (clearly). 

Now this also needs the same branch names as that here but in MAPL.

mathomp4 (approved):

PR #279, opened 2024-06-20
Revert back to working version of Util and updated SST to latest

rtodling:
The version of GEOS_Util in 5.29.5-p6 has troubles: 

1) monthly plots don't work
2) SST updates are incorrect 

These are now reversed back:

1) v1.0.2 - which has been used in 5.30.2 and has not issues
2) brought in the latest changes in the SST handles (from v2.0.8)

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -5,7 +5,7 @@ GEOSadas:
 env:
   local: ./@env
   remote: ../ESMA_env.git
-  tag: v3.4.0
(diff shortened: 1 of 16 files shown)


mathomp4 (approved):

mathomp4:
I guess this branch is old enough that it was before I put CircleCI in. Luckily it doesn't seem to cause an issue with merging, though if you like, I could put in a dummy config that does nothing so it's not annoying.

PR #288, opened 2024-10-15
Modify twindow.rc, twindow_m2.rc to add (excluded) newer wind types

gmao-msienkie:
Some new wind types (NC005072, NC005081) that are not used.
This will avoid messages in the output about these unknown wind types.
There should be no difference in the output.


gmao-msienkie:
Tests with the modified rc files gave identical results in BUFR output.  These messages no longer printed:
 twindow: error, NC005081 not found in subset table
 twindow: error, NC005072 not found in subset table

rtodling:
@gmao-msienkie I believe these changes are mainly aimed at R21C (M21C), am I correct? It seems they would be ok going into develop - but these would need to be brought in to R21C. Can you confirm, pls?

gmao-msienkie:
These changes are actually only relevant for the older systems like MERRA2 and FPIT.  It is run as a data preprocessing step in operations.  We wanted to limit the amount of wind data read in to avoid overflow of the internal array, and to keep the amount of AMVs assimilated similar to what had been using.
We do have this code in Github however, inherited from the older CVS repository.  
Someone else will need to take on maintenance of this code since it seems like there is interest in keeping MERRA2 going for longer than I anticipated.  It is possible it may need to be modified for other changes in the AMVs before MERRA2 is finally retired.   
R21C and modern systems do not need this although there will still need to be a process of updating R21C when changes occur to the AMV input - which I will happily leave to those who will be maintaining it.

rtodling (approved):
Thanks for these changes.

PR #311, opened 2024-11-26
M21C DAS update 2

sdrabenh:
Changes proposed by @elakkraoui for `GEOSadas-5.30.3-M21C-6`. The main changes are increase aircraft max tail, turn on GOES18, add avhrr to pcscaling.

Marking as 0-diff relative to GCM. I suspect most of these changes are 0-diff to the ADAS except for the obs changes.

@rtodling please review and tag at your earliest convenience. Many thanks.


rtodling (approved):
I will take these changes in, but there are more changes that need to be made on top of these - these are not enough since the analysis is missing the OMPS-LP fixes and the because AOD analysis still - and pyods in GMAO_Shared is out of sync.

elakkraoui:
@rtodling, I didn't include OMPS-LP fixes because your PR was already merged in. Did I miss something? As for the AOD analysis, I don't get what you mean by out of sync. Are you referring to something introduced with the Milan tag?

PR #322, opened 2025-01-10
Update "offline" aircraft bias to use dynamic allocation of tails.

gmao-msienkie:
Note - this is the same method of calculating aircraft bias coefficients as in MERRA2, but the format of the coefficient files was changed to be compatible with the aircraft varBC coefficient format used in the current GEOSadas.  

We are adding dynamic allocation to the tail-based coefficient arrays in MERRA2.  The corresponding code changes have been copied into this branch.

While it isn't likely that this code will be used in the current GEOSdas it is being maintained in case someone wants to run the MERRA2 bias correction in the current GEOSdas.


rtodling (approved):
Thanks for keeping an eye on this Meta.

PR #328, opened 2025-02-11
Changes to allow 'radcor' programs to work with enhanced 255-level raobs

gmao-msienkie:
Recently NCEP modified the 'prepbufr' processing to include enhanced radiosonde profiles derived from high-density radiosonde data.  These soundings may have up to 255 wind and mass levels.  Tests with a sample file caused MERRA2 pre-processing to fail.  Increasing the maximum number of elements in sounding arrays to a larger number allowed the program to run.  Using 512 in place of 255 would handle things if 255 mass + 255 wind levels are needed.  Using an array dimensioned 512 in a call to read a BUFR field with dimension 255 is okay since the UFBINT call will return the number of elements used in the array.


rtodling (approved):
Thank you

PR #329, opened 2025-02-26
Update documentation

gmao-msienkie:
Added a description of the format of aircraft bias coefficient tables to GMAO_Etc/GMAO_bias (README.aircraft_bias).
Added a note about the 'gmao_global_tlapmean.rc' values based on the old variational bias correction so the entries are scaled by 0.01 from what is used in the current radiance bias correction.


rtodling (approved):
Thank you

PR #341, opened 2025-03-18
Remove possibly corrupt file before GCM runs

rtodling:
When model ensemble member fails, the model leaves behind some of its history output. Typically, if the history output is sane, a resubmission of the ensemble successfully completes the previously failed run. At times, however, some of the history output might be corrupt, in these cases, a resubmission of the ensemble will continue to fail since logics in MAPL will want to open the existing file - however, for corrupt files this causes the model to fail .. and it will fail over and over again. 
Changing the MAPL logics to check on the sanity of files it tries to open is not so simple.

This PR introduces a script logic that removes corrupt files before the model starts running. This has been tests on a couple of failed cases and it seems to work like a charm.


rtodling:
@mathomp4 This is odd! A one file change and the building fails.  Is there a way to restart the CI - have it try again?

mathomp4:
Ohhh. The CI might need updates, I think it's finding too old of a Docker image. For now I'll approve and look at updating it if I can.

Trust builds on discover more than CI for now :)

mathomp4 (approved):

rtodling:
Thank you!

PR #346, opened 2025-03-21
Update the runjob and checkinput perl scripts to set $ESMABIN equal to …

gmao-jstassi:
… their directory location rather than using a hard-coded value. This will allow the scripts to work correctly if the build's install directory location is changed.

An option was added to the runjob script to call checkinput. This gives more visibility to the checkinput script.


rtodling:
@gmao-jstassi I have just updated this PR to be in sync with the latest  in develop.

rtodling (approved):
Thank you Joe.

PR #347, opened 2025-04-01
Add option for setting 'gross check' value to init_coeffs.f90

gmao-msienkie:
Add command line option '-L' to specify the value of the gross check limit to use instead of the default value of 200K.


rtodling (approved):
Seems good by me.

PR #327, opened 2025-01-27
update setup procedures for LADAS 

saraqzhang:
for integrated setup of adas and ldas, modifications and updates are implemented in fvsetup and several scripts and rc files  that are used in the fvsetup procedure.  the modifications to other related external components and ldas are updated by pull requests to the corresponding repositories. 

Related PR: https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/81 (merged; follow-up PR in prep)

Requires:
https://github.com/GEOS-ESM/GEOSgcm_App/pull/689 (merged)
https://github.com/GEOS-ESM/GMAO_Shared/pull/370 (merged)

PR #352, opened 2025-05-01
additional modifications in fvsetup for more integrated land-atmosphere DAS setup and configuration

saraqzhang:
Additional modifications of fvsetup for more integrated land-atm DAS setup and configuration (on top of #327). 

Contingent on: 
- https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/94 (merged)
- https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/102 (merged) 
- MAPL v2.57.0 (for GEOSldas_GridComp v3.1.0; also satisfies ExtData fix of v2.56.1 required for GEOSadas)


saraqzhang:
@gmao-rreichle    changes here corresponding to ldasGC pr#94 draft.

src/Applications/GEOSdas_App/fvsetup
```diff
@@ -579,21 +579,22 @@ $ENV{"PATH"} = $FindBin::Bin .":$ENV{PATH}";
...
 # Call LDAS setup script.
-#  Provide LDAS config input files in $ldassetup/: det_exeinp.txt, det_batinp.txt. 
-#  For Hyb-4dEnVar, do same for ens_exeinp.txt, ens_batinp.txt. 
+#  Provide LDAS config input files in $ldassetup/: ldas_exeinp.txt, ldas_batinp.txt. 
 #  See GEOSldas/README.md for generating sample config input files.
 # ------------------------------------------------------------------------
     print "\n checkprintout  ldas_ana == 1 ($ldas_ana)\n";
     if ($ldas_ana == 1 ) {
-        $cmd = "$fvbin/ldas_setup setup $fvhome/lana --account $gid --nymdb $nymdb --nhmsb $nhmsb $ldassetup/det_exeinp.txt $ldassetup/det_batinp.txt";
+        $cmd = "$fvbin/ldas_setup setup $fvhome/lana --account $gid --nymdb $nymdb --nhmsb $nhmsb --ladas_cpl 1 --agcm_res $res --rstdir $ldasics $ldassetup/ldas_exeinp.txt $ldassetup/ldas_batinp.txt";
```
gmao-rreichle: Should we call ldas_setup with the "-v" (--verbose) option?  Not sure if this would/should be captured in the ADAS setup log, assuming such a log exists.

gmao-rreichle: The variable $res seems to exist in the desired format elsewhere in fvsetup as $agcm_im4.  Could we use $agcm_im4 here to avoid the string modifications inside ldas_setup? 
There is also another variable, $bcsresa, that seems to hold the same info. 
Also, what we are really interested in here is the BCS info.  The BCS info consists of the AGCM grid & resolution as well as the version.  From $res alone we can't identify the bcs that need to be used in LDAS.

src/Applications/GEOSdas_App/fvsetup
```diff
@@ -579,21 +579,22 @@ $ENV{"PATH"} = $FindBin::Bin .":$ENV{PATH}";
...
         print "$cmd\n";
         $rc = system $cmd;
         die "Failed to setup LDAS for det, check LDAS config input files $!" if ( $rc );
         chomp($aens_replay);
         if (($hyb_ens == 4) && ($aens_replay eq "no")) { 
-           $rc = system("$fvbin/ldas_setup setup $fvhome/lana --account $gid --nymdb $nymdb --nhmsb $nhmsb $ldassetup/ens_exeinp.txt $ldassetup/ens_batinp.txt");
-           die "Failed to setup LDAS for ens, check LDAS config input files $!" if ( $rc );
-       }
+	    $cmd = "$fvbin/ldas_setup setup $fvhome/lana --account $gid --nymdb $nymdb --nhmsb $nhmsb --ladas_cpl 2 --agcm_res $aens_res --rstdir $ldasics4ens $ldassetup/ldas_exeinp.txt $ldassetup/ldas_batinp.txt";
```
gmao-rreichle: The comments above about -v, $res, and bcs version also apply to the setup of the LDAS for the ensemble component of the ADAS.

gmao-rreichle (commented):
@saraqzhang, please double-check my edits in the most recent commit https://github.com/GEOS-ESM/GEOSadas/pull/352/commits/03c1dda92c60ea5361853a334dd05a4f982a0583.
Also see inline comments below.

src/Applications/GEOSdas_App/fvsetup
saraqzhang: can add -v to save printouts in LOG .

saraqzhang: modify this  to use $agcm_im4  and  a constructed $aens_im4 as inputs to ldas_setup, and  modify ldas_setup accordingly.

src/Applications/GEOSdas_App/fvsetup
saraqzhang: will need to  have a closer look at BCs version.

src/Applications/GEOSdas_App/fvsetup
saraqzhang: will take a closer look at BCs version.

weiyuan-jiang: It is cleaner to put all non-positional params after positional ones

saraqzhang:
updates are tested in setup of a hybrid4dnvar LADAS  ( with ldas_setup updated in ldasGC PR#94)

saraqzhang:
@gmao-rreichle   update command line inputs corresponding to ldasGC PR#94.  For the path info of ldas SPEC NML and MWRTM param,  the options of default and user input to fvsetup  are also added.

saraqzhang:
@gmao-rreichle  note that the default path currently I put in should be treated as a place holder, we should have "official"  location and files that contain most updated specifications  ( for instance in spec nml that you pointed to Jana yesterday)  before the PR to be merged.

gmao-rreichle:
Draft list of open items for discussion with Ricardo & Joe:
- Can ldas_setup obtain the number of ens members from fvsetup?
- How should we set the SLURM parameters for the LDAS (walltime, n-tasks-per-node, n-tasks, ...)?
- Where should we keep the MWRTM parameter files?
- Where should we keep the LDAS NML input files? 
- Change fvsetup to use BCS in new layout (as does gcm_setup for v11.7 model)?  Then we would not need to hardwire the BCS_PATH into ldas_setup.

saraqzhang:
@gmao-rreichle  last commit 8122422 for ladas setup  by runjob ( fvsetup). The system is built with adas v5.42.4,  ldasGC branch feature/saraqzhang/setupfunc4ladas ( updated from v.2.0.0) ;  MAPL  v2.55.0 ( updated from v2.52.0);  and GEOS_Util  g5.42.0 .  The  setup test is successful.   A coupled  hybrid exp run test is on going.

src/Applications/GEOSdas_App/fvsetup
```diff
@@ -590,21 +590,21 @@ $ENV{"PATH"} = $FindBin::Bin .":$ENV{PATH}";
...
 
 # Call LDAS setup script.
-#  Provide LDAS config input files in $ldassetup/: det_exeinp.txt, det_batinp.txt. 
-#  For Hyb-4dEnVar, do same for ens_exeinp.txt, ens_batinp.txt. 
-#  See GEOSldas/README.md for generating sample config input files.
 # ------------------------------------------------------------------------
     print "\n checkprintout  ldas_ana == 1 ($ldas_ana)\n";
     if ($ldas_ana == 1 ) {
-        $cmd = "$fvbin/ldas_setup setup $fvhome/lana --account $gid --nymdb $nymdb --nhmsb $nhmsb $ldassetup/det_exeinp.txt $ldassetup/det_batinp.txt";
+        $cmd     = "$fvbin/ldas_setup setup $fvhome/lana exeinp batinp -v --account $gid --nymdb $nymdb --nhmsb $nhmsb --ladas_cpl 1 --agcm_res $agcm_im4 --varwindow $varwindow --bcs_version $landbcs --rstloc $ldasics --LDAS_nml_path $ldasnml_path --mwrtm_path $ldasrtm_path";
```
gmao-rreichle:
```suggestion
        $cmd     = "$fvbin/ldas_setup setup $fvhome/lana dummy_exeinp dummy_batinp -v --account $gid --nymdb $nymdb --nhmsb $nhmsb --ladas_cpl 1 --agcm_res $agcm_im4 --varwindow $varwindow --bcs_version $landbcs --rstloc $ldasics --LDAS_nml_path $ldasnml_path --mwrtm_path $ldasrtm_path";
```

src/Applications/GEOSdas_App/fvsetup
```diff
@@ -590,21 +590,21 @@ $ENV{"PATH"} = $FindBin::Bin .":$ENV{PATH}";
...
         print "$cmd\n";
         $rc = system $cmd;
         die "Failed to setup LDAS for det, check LDAS config input files $!" if ( $rc );
         chomp($aens_replay);
         if (($hyb_ens == 4) && ($aens_replay eq "no")) { 
-           $rc = system("$fvbin/ldas_setup setup $fvhome/lana --account $gid --nymdb $nymdb --nhmsb $nhmsb $ldassetup/ens_exeinp.txt $ldassetup/ens_batinp.txt");
-           die "Failed to setup LDAS for ens, check LDAS config input files $!" if ( $rc );
-       }
+            $aens_im4 = sprintf "%04i", $aens_im ;
+	    $cmd = "$fvbin/ldas_setup setup $fvhome/lana exeinp batinp -v --account $gid --nymdb $nymdb --nhmsb $nhmsb --ladas_cpl 2 --agcm_res $aens_im4 --varwindow $varwindow --bcs_version $landbcs --rstloc $ldasics4ens --LDAS_nml_path $ldasnml_path --mwrtm_path $ldasrtm_path";
```
gmao-rreichle:
```suggestion
	    $cmd = "$fvbin/ldas_setup setup $fvhome/lana dummy_exeinp dummy_batinp -v --account $gid --nymdb $nymdb --nhmsb $nhmsb --ladas_cpl 2 --agcm_res $aens_im4 --varwindow $varwindow --bcs_version $landbcs --rstloc $ldasics4ens --LDAS_nml_path $ldasnml_path --mwrtm_path $ldasrtm_path";
```

gmao-rreichle (commented):
@saraqzhang : I suggest making it more explicit that "exeinp" and "batinp" in the call to ldas_setup are dummy strings.  See suggested changes below.  I think you can "accept/merge" the changes directly on GitHub.com using the buttons in the comments boxes.

saraqzhang:
@gmao-rreichle    the edits are merged.

gmao-rreichle:
> @gmao-rreichle the edits are merged.

@saraqzhang, for some reason only one of the two edits had been merged. I just merged the other edit.  I also updated the status of GEOSldas_GridComp PR#94, which has been merged, and removed the "contingency" label.

src/Applications/GEOSdas_App/fvsetup
```diff
@@ -3540,30 +3540,50 @@ sub set_ldasANA {
...
-        $ldassetup = query("  LDAS setup input path?", $dflt);
-      
+ 
+        $dflt        = "/discover/nobackup/$user/ldas_exp_C360/";
+        $prompt      = "  Location of LDAS initial conditions for deterministic component? (full_path/ldas_exp) ";
+        $ldasics4det = query($prompt, $dflt);
+           die ">>> Error <<< Cannot find directory, $ldasics4det;"
+                unless -d $ldasics4det;
+
+	# FOR NOW, ASK USER FOR NUMBER OF ENS MEMBERS; COULD DETERMINE FROM ICS DIR
```
gmao-rreichle: @saraqzhang : Maybe we should just obtain ldas_nens_det from the files in ldasics4det?

src/Applications/GEOSdas_App/fvsetup
```diff
@@ -3540,30 +3540,50 @@ sub set_ldasANA {
...
+            $dflt        = "/discover/nobackup/$user/ldas_exp_C90/";
+            $prompt      = "  Location of LDAS initial conditions for ens? (full_path/ldas_exp) ";
+            $ldasics4ens = query($prompt);
+            die ">>> Error <<< Cannot find directory, $ldasics4ens;"
+                unless -d $ldasics4ens;
+
+	    # FOR NOW, ask user for number of members in *atm* ensemble b/c setup of atm ens requires I/O
+	    # (and in any case users may specify "later" setup of atm ens)
+	    # COULD CHANGE TO DETERMINING NUMBER OF ENS MEMBERS FROM ldasics4ens AND THEN VERIFY LATER
+	    # AGAINST NUMBER OF ATM ENS MEMBERS
```
gmao-rreichle: @saraqzhang : Maybe we should obtain ldas_nens_atmens from the files in ldasics4ens.  Then we can keep confirm later that ldas_nens_atmens matches the number of atm ens members (after the setup of the atm ens has happened)

gmao-rreichle (commented):
@saraqzhang : I made some tentative changes to fvsetup so it should work with the latest changes in ldas_setup.  There are probably bugs in the implementation.  I hope the overall approach makes sense.

src/Applications/GEOSdas_App/fvsetup
saraqzhang: we can do that. count the members in rs/ .

src/Applications/GEOSdas_App/fvsetup
saraqzhang: we can find the number in ldas ics and compare/confirm with the one from fvsetup.

saraqzhang:
@gmao-rreichle commit# 54dabcc  add function to find nens from atmens, ldas restart 4det, ldas_restart 4ens.  nens pass to ldas_setup via commandline.  It is tested by runjob ( fvstup calling ldas_setup ).

src/Applications/GEOSdas_App/fvsetup
```diff
@@ -3874,6 +3862,37 @@ EOF
...
 }
 
+#======================================================================
+sub get_nens { 
+    #default 
+    $atm_nens = 32;
+    $ldas_nens_det = 24;
+    $ldas_nens_atmens = $atm_nens; 
+    
+    @enslist = (<$fvhome/atmens/mem???>);
```
gmao-rreichle: If we call get_nens() at the time we call ldas_setup, then this list is presumably empty because the atm ensemble  has not yet been set up.  Then $atm_ens remains at the default value and the check becomes pointless.  Don't we need to count the atm ens members **after** the atm ens has been set up?   (allowing for the fact that the user may select "later" and the atm ens isn't set up by fvsetup)

src/Applications/GEOSdas_App/fvsetup
```diff
@@ -3874,6 +3862,37 @@ EOF
...
+    if ( scalar(@enslist) > 1 ) {
+       $atm_nens = scalar(@enslist);
+    }
+    $ldasdomain ="CF$agcm_im4x6C_GLOBAL";
```
gmao-rreichle: Does this work reliably?  I think it would probably be safer to use:
`$ldasdomain = "CF" . $agcm_im4 . "x6C_GLOBAL";`
or
`$ldasdomain = "CF${agcm_im4}x6C_GLOBAL";`
Also, does this create a global variable?  Do we really need to have $ldasdomain available throughout the package?  I think prefacing by "my" will limit the scope, as in:
`my $ldasdomain = "CF" . $agcm_im4 . "x6C_GLOBAL";` 
Sorry if I'm getting this all wrong.  I don't really know perl.

src/Applications/GEOSdas_App/fvsetup
```diff
@@ -3874,6 +3862,37 @@ EOF
...
+    $ldasics_det = "$ldasics/output/$ldasdomain";
```
gmao-rreichle: I think this needs to be replaced with:
`$ldasics_det = "$ldasics4det/output/$ldasdomain";`
I changed the name $ldasics to $ldasics4det in a recent commit for clarity.
This may also suggest that the testing of the latest revisions failed in the sense that the defaults kicked in, and the actual counting never happened.  (If perl doesn't complain if $ldasics is unknown and produces an empty @ens_list, we probably end up with $ldas_nens_det falling back on the default value.  I think perl at best produces a warning in this case and you end up with empty strings, but not sure.)

src/Applications/GEOSdas_App/fvsetup
```diff
@@ -3874,6 +3862,37 @@ EOF
...
+    @enslist = (<$ldasics_det/rs/ens????>);
+    if ( scalar(@enslist) > 1 ) {
+        $ldas_nens_det = scalar(@enslist);
+    }
+    $aens_im4 = sprintf "%04i", $aens_im ; 
+    $ldasdomain ="CF$aens_im4x6C_GLOBAL";
```
gmao-rreichle: see comment above about safer string concatenation

src/Applications/GEOSdas_App/fvsetup
```diff
@@ -3874,6 +3862,37 @@ EOF
...
+    $ldasics_ens = "$ldasics4ens/output/$ldasdomain";
+    @enslist = (<$ldasics4ens/rs/ens????>);
+    if ( scalar(@enslist) > 1 ) {
+       $ldas_nens_atmens = scalar(@enslist);
+    }
+    if ( $atm_nens ne  $ldas_nens_atmens ) {
```
gmao-rreichle: this check probably doesn't really do anything meaningful unless we run get_nens() after the atm ens has been set up (because $atm_nens is set to its default value)

src/Applications/GEOSdas_App/fvsetup
```diff
@@ -3874,6 +3862,37 @@ EOF
   }
 }
 
+#======================================================================
+sub get_nens {
```
gmao-rreichle: I think the function name should be more specific, something like "get_nens4ldas()". Otherwise it's very unclear that the only purpose of the function is to count the number of ens member for the purpose of ldas_setup

src/Applications/GEOSdas_App/fvsetup
```diff
@@ -3548,7 +3549,7 @@ sub set_ldasANA {
             die ">>> Error <<< Cannot find directory, $ldashold;"
                 unless -d $ldashold;
 
-        @bkglfo_files = (<$ldashold/*bkg.lfo*${nymdb}_${hhb}*>);
+        @bkglfo_files = (<$ldashold/*bkg*lfo*${nymdb}_${hhb}*>);
```
gmao-rreichle, line 3553: I noticed that in this line you changed `*bkg.lfo*` (with a dot in the middle) back to `*bkg*lfo*`.  Since we're looking for files whose names include "bkg.lfo", why can't we use the more specific string?

gmao-rreichle (commented):
Hi @saraqzhang : Thanks for making the change to count the # of ens members for ldas_setup here in fvsetup.  I have a few questions/suggestions re. the revisions, see comments below

src/Applications/GEOSdas_App/fvsetup
saraqzhang: there are two possible file names here , one is the tarr file  from adas archive ( bkglfo), the other one is a collection that user may provide ( bkg.lfo,  *nc4) . so * bkg*lfo* will serve for both. 
I think in the next revision to have bootstrap for ldas we will remove the option of user providing a collection. If adas ICS archive don't have bkglfo tar in rs/ ,  ldas will be in bootstrap.

src/Applications/GEOSdas_App/fvsetup
saraqzhang: @gmao-rreichle  I consulted Joe and also tested fvsetup,  the top perl script actually waits for the spun-off perl script that installs atmens/ to finish before going to next action.  SInce that script is called before ldas_setup, by the time we call ldas_setup we have atmens/  in fvhome, thus we can get the nens then.

src/Applications/GEOSdas_App/fvsetup
saraqzhang: ldasdomain is a temp variable, not needed anywhere else but here to hold the string we put together.  technically we can incorporate the " ..."  ( without $ldasdomain)  in the path, but I think it looked messier.  $ldasdomain worked in the test, of course we can make a safer version.

src/Applications/GEOSdas_App/fvsetup
saraqzhang: the test passed.  see comment before:  "the top perl script actually waits for the spun-off perl script that installs atmens/ to finish before going to next action. SInce that script is called before ldas_setup, by the time we call ldas_setup we have atmens/ in fvhome, thus we can get the nens then."

src/Applications/GEOSdas_App/fvsetup
saraqzhang: see comment above.

src/Applications/GEOSdas_App/fvsetup
saraqzhang: sure we can rename it.

src/Applications/GEOSdas_App/fvsetup
gmao-rreichle: Ok, thanks.  I had assumed the tar file retains the exact string from the nc4 files (i.e., something like "bkg.lfo*.tar").  Does the ADAS create this tar file for archiving and later use?  If so, we could presumable change the name of the tar file so the dot is used consistently, although this may be inconsistent with the naming convention of the other bkg files...  It's not worth worrying about any further.  I'll resolve the comment.

src/Applications/GEOSdas_App/fvsetup
gmao-rreichle: So if the top perl script waits, we are expecting to have the atmens installed.  In this case, we should initialize all N variables to -9999 or something.  Otherwise we would get past the check (by falling back on the default values) even if something funny happened with the atmens installation.

src/Applications/GEOSdas_App/fvsetup
saraqzhang: will make the change.

src/Applications/GEOSdas_App/fvsetup
saraqzhang: will make the change.

src/Applications/GEOSdas_App/fvsetup
saraqzhang: will make the change

saraqzhang:
@gmao-rreichle  modifications are made ( commit 484b861) and runjob (fvstup  -> ldas_setup)  test passed.

saraqzhang:
@gmao-rreichle   the last commit did not consider the  scenario of "later"  for atmens ICS. I will work on that  later today.

saraqzhang:
@gmao-rreichle c2a59da deal with atm_ens input with "later " option for atmens  ICS installation

gmao-rreichle:
> @gmao-rreichle [c2a59da](https://github.com/GEOS-ESM/GEOSadas/commit/c2a59da4e361b58c2b1ad82d8e3b2b3d3f1f48e3) deal with atm_ens input with "later " option for atmens ICS installation

@saraqzhang : Thanks for adding this. I edited the query text a bit and fixed an error in the variable name (I think) https://github.com/GEOS-ESM/GEOSadas/pull/352/commits/d4480d0efb5a4ce456c42f95a98f800e252f872b.  I also updated the branch to the latest develop.  Please double check.

saraqzhang:
yes that was a typo. d4480d0 fixed it.

gmao-rreichle (dismissed):

gmao-rreichle (dismissed):

gmao-rreichle (commented):
Sorry, my approval was too early.  Need to create a new GEOSldas_GridComp release and update the components.yaml file accordingly.  Will do ASAP

saraqzhang:
@gmao-rreichle @rtodling  the current system ( v5.42.11 + PR#352 ) passed tests of  built,  LADAS setup and  2-day hybrid cycling run. a tag is created LADAS_v5.42.11 .

gmao-rreichle (dismissed):

rtodling:
Is the change in MAPL a zero-diff change?

weiyuan-jiang:
> Is the change in MAPL a zero-diff change?

Between MAPL v2.52 and v2.57, there is no mention about non-zero-diff.  So it should be zero-diff

gmao-rreichle:
> > Is the change in MAPL a zero-diff change?
> 
> Between MAPL v2.52 and v2.57, there is no mention about non-zero-diff. So it should be zero-diff

Yes, it's 0-diff.  @rtodling, aren't you already using  MAPL v2.56.1 (which includes the ExtData fix) in your branch?  v2.57.0 only changes EASE-grid routines, which is 0-diff (and the EASE grid isn't used by ADAS anyway).  But as @weiyuan-jiang said, MAPL should be 0-diff between v2.52 and v2.57

weiyuan-jiang (dismissed):

rtodling:
Sara and @gmao-rreichle Just letting you know here that I pulled these changes yesterday and tried testing them, but unfortunately, I am unable to corroborate them being zero-diff. I believe the issue is in MAPL, but I have no idea why.  I am getting in touch with the SI group so we can go over the differences between the two versions of MAPL in question:

https://github.com/GEOS-ESM/MAPL/compare/v2.52.0...v2.57.0

gmao-rreichle:
> Sara and @gmao-rreichle Just letting you know here that I pulled these changes yesterday and tried testing them, but unfortunately, I am unable to corroborate them being zero-diff. I believe the issue is in MAPL, but I have no idea why. I am getting in touch with the SI group so we can go over the differences between the two versions of MAPL in question:
> 
> [GEOS-ESM/MAPL@v2.52.0...v2.57.0](https://github.com/GEOS-ESM/MAPL/compare/v2.52.0...v2.57.0)

@rtodling : I'm having a hard time believing that the changes in MAPL caused any non-trivial changes.   As @weiyuan-jiang said above, MAPL v2.52 and v2.57 should be zero-diff (https://github.com/GEOS-ESM/GEOSadas/pull/352#issuecomment-3070274063).  Having said that, I did find the following comment in the MAPL v2.57 release notes (https://github.com/GEOS-ESM/MAPL/releases/tag/v2.57.0):

`NOTE: There is one change that can affect some diagnostic prints made by MAPL. The per-step diagnostic print was changed from being hardcoded as AGCM Date to now trigger off of the ROOT_NAME in CAP.rc. So, if ROOT_NAME is GEOSldas, the print will be GEOSldas Date instead of AGCM Date.`

Could the non-zero-diff issue that you're finding be related to something like that?

rtodling:
> > Sara and @gmao-rreichle Just letting you know here that I pulled these changes yesterday and tried testing them, but unfortunately, I am unable to corroborate them being zero-diff. I believe the issue is in MAPL, but I have no idea why. I am getting in touch with the SI group so we can go over the differences between the two versions of MAPL in question:
> > [GEOS-ESM/MAPL@v2.52.0...v2.57.0](https://github.com/GEOS-ESM/MAPL/compare/v2.52.0...v2.57.0)
> 
> @rtodling : I'm having a hard time believing that the changes in MAPL caused any non-trivial changes. As @weiyuan-jiang said above, MAPL v2.52 and v2.57 should be zero-diff ([#352 (comment)](https://github.com/GEOS-ESM/GEOSadas/pull/352#issuecomment-3070274063)). Having said that, I did find the following comment in the MAPL v2.57 release notes (https://github.com/GEOS-ESM/MAPL/releases/tag/v2.57.0):
> 
> `NOTE: There is one change that can affect some diagnostic prints made by MAPL. The per-step diagnostic print was changed from being hardcoded as AGCM Date to now trigger off of the ROOT_NAME in CAP.rc. So, if ROOT_NAME is GEOSldas, the print will be GEOSldas Date instead of AGCM Date.`
> 
> Could the non-zero-diff issue that you're finding be related to something like that?

@gmao-rreichle What you saying above might be an issue - but only when LDAS is on, which in my testing it isn't. 

I know what the issue is: the AOD analyses are not being processed correctly. Now what that is the case is still under investigation. I asked for Matt and Atanas's help - I am re-running my test case to get a dump of the yaml files to see what might be the issue.

It is possible that when Sara made the test, her baseline was not using the AOD analysis correctly - many people overlook that - which means the run might have zero-increment for AOD, in which case, she would have gotten a zero-diff result from the before and after changes here.

gmao-rreichle:
@rtodling : I can't speak to what exactly Sara tested and what you may be testing.  But taking a step back, I can't possibly see how this PR has anything to do with the AOD analysis.  There are only 2 files changed in this PR:

- `components.yaml` impacts the MAPL version, which should be zero-diff (unless the documentation by the SI team is wrong). 
- `fvsetup` modifies the LDAS setup within the ADAS setup.  If you're not running the LDAS, this should have no impact at all, and certainly not on the AOD analysis.  I just went through the fvsetup changes again.  The majority of the changes are within an "lana==.true." block and an LDAS-relates sub function.  The only change that is not so constrained has to do with sequencing the setup of the atmos ensemble and the LDAS setup: https://github.com/GEOS-ESM/GEOSadas/blob/3dd73e34c5ca48627a666db4a443bdd9c1012454/src/Applications/GEOSdas_App/fvsetup#L3858  If the system sets up and runs, how could the changes in this PR cause non-zero diff results?

rtodling:
> @rtodling : I can't speak to what exactly Sara tested and what you may be testing. But taking a step back, I can't possibly see how this PR has anything to do with the AOD analysis. There are only 2 files changed in this PR:
> 
> * `components.yaml` impacts the MAPL version, which should be zero-diff (unless the documentation by the SI team is wrong).
> * `fvsetup` modifies the LDAS setup within the ADAS setup.  If you're not running the LDAS, this should have no impact at all, and certainly not on the AOD analysis.  I just went through the fvsetup changes again.  The majority of the changes are within an "lana==.true." block and an LDAS-relates sub function.  The only change that is not so constrained has to do with sequencing the setup of the atmos ensemble and the LDAS setup: https://github.com/GEOS-ESM/GEOSadas/blob/3dd73e34c5ca48627a666db4a443bdd9c1012454/src/Applications/GEOSdas_App/fvsetup#L3858
>     If the system sets up and runs, how could the changes in this PR cause non-zero diff results?

@gmao-rreichle this PR relies on a change in MAPL that brings in a ton of stuff related to ExtData2G which has nothing to do with LDAS but has a large pontential to cause troubles. The problem is not in the LDAS-related knobs and changes; the problem is in MAPL and how it changes something about handling the info in the yaml files that drive the model.

rtodling:
> > @rtodling : I can't speak to what exactly Sara tested and what you may be testing. But taking a step back, I can't possibly see how this PR has anything to do with the AOD analysis. There are only 2 files changed in this PR:
> > 
> > * `components.yaml` impacts the MAPL version, which should be zero-diff (unless the documentation by the SI team is wrong).
> > * `fvsetup` modifies the LDAS setup within the ADAS setup.  If you're not running the LDAS, this should have no impact at all, and certainly not on the AOD analysis.  I just went through the fvsetup changes again.  The majority of the changes are within an "lana==.true." block and an LDAS-relates sub function.  The only change that is not so constrained has to do with sequencing the setup of the atmos ensemble and the LDAS setup: https://github.com/GEOS-ESM/GEOSadas/blob/3dd73e34c5ca48627a666db4a443bdd9c1012454/src/Applications/GEOSdas_App/fvsetup#L3858
> >   
> >   If the system sets up and runs, how could the changes in this PR cause non-zero diff results?
> 
> @gmao-rreichle this PR relies on a change in MAPL that brings in a ton of stuff related to ExtData2G which has nothing to do with LDAS but has a large pontential to cause troubles. The problem is not in the LDAS-related knobs and changes; the problem is in MAPL and how it changes something about handling the info in the yaml files that drive the model.

btw: you can be perplexed, but your can't deny the fact that I am inheriting a bug since I see it in the results.

gmao-rreichle:
@rtodling : Sorry, I didn't mean to imply that you're not dealing with a bug.  Clearly, you are seeing non-zero-diff results.  What I'm saying is just that the bug can't be related to the changes in the PR -- with the exception of the MAPL change. It may well be that the MAPL change is causing problems for the ADAS, and that this hasn't been properly tested by Sara.  In my comments, I was going by the MAPL release documentation, which seems to suggest that the change was zero-diff.  But perhaps this testing by the SI team doesn't include a full ADAS test.  

In any case, this suggests a bigger issue.  This PR needs the MAPL update because the newer MAPL includes new features that are needed by the LDAS.  These features aren't related to what's causing your problem.  Going forward, we probably have to choose between the following:
1. Make ADAS work with the newer MAPL versions (which I assume has to happen eventually), or
2. ask the SI team to create a patch to MAPL v2.52 that includes what the LDAS needs (but excludes the offending ExtData2G changes), or
3. not merge this PR (#352).

You can probably guess that I'd love to avoid option 3.  It would set back the LDAS integration even further.  

I can't speak to option 1 and how relevant the non-zero-diff changes are. 

Maybe it's worth inquiring with @mathomp4 about option 2?  I'm somewhat pessimistic about this being an easy fix, but it shouldn't hurt to ask.

rtodling:
@saraqzhang @gmao-rreichle just to keep you abreast, I am able to identify the changes that went into MAPL that cause the issue I'm having. I am working with the SI group to come up with/ a meaningful fix. The error has nothing to do with LDAS stuff but with how MAPL handles the AOD analyses file and the assimilaiton of aerosol info.

rtodling:
Now with MAPL v2.57.1 things work and give me zero-diff. Thanks to SI Team for patching MAPL.

rtodling (approved):
The patch to MAPL 2.57 gets things to work fine.

PR #353, opened 2025-05-01
Add cycling JEDI capability to GEOS

rtodling:
This introduces JEDI-capability to GEOS. 

A comprehensive Wiki is being provided to guide users into exercising the new feature:

https://github.com/GEOS-ESM/GEOSadas/wiki/Y.-GEOS%E2%80%90JEDI

A simplified GSI configuration is introduced to allow for fair comparison with JEDI since the latter does not yet have all feature of the former.

This is zero diff to the regular ADAS.


mathomp4 (approved):
Trust in @rtodling

PR #355, opened 2025-05-02
Adding 4d capability for JEDI 

rtodling:
This is simply the knobs - 4d is not yet ready to be exercised - esp since the version of JEDI this is pegged to is not quite ready.

But these are zero diff to previous results.


mathomp4 (approved):

rtodling:
Thanks Matt.

PR #364, opened 2025-05-05
update 'twindow' resource files to remove restriction on GOES-19

gmao-msienkie:
It turned out that it wasn't necessary to restrict the GOES-19 AMV from time windowed AMV files.
The GOES-18 and GOES-19 data did not overlap.


rtodling (approved):
Thanks for this Meta.

PR #373, opened 2025-05-21
Feature/rtodling/fv3 jedi 1335

### LINKED ISSUE #161: Updates to 5_29_4 tag

### LINKED ISSUE #198: Update GEOSdas.csm to handle local acquire for gaas_obs
The GEOSdas.csm file is modified to handle local acquire of gaas_obs data using the entries in the obsys-gaas.rc file.

### LINKED ISSUE #232: Das script will need to be updated to handle ExtData2G
@rtodling 
This is a public service announcement based on expected changes coming to the model.

All the chemistry codes are in the process of transitioning to use the new version of ExtData which uses yaml files rather than rc file. This is just a heads up that the DAS scripting will undoubtedly need to be updated to handle this. But is gets less complicated not more!

For now both versions (the rc and yaml files) exists for testing (ExtData2G itself is turned on by a flag in the CAP.rc but when all testing is complete ExtData2G will become the default.

The main pertinent change for the DAS scripting is
1. All the magic date logic can go away, no more of this awful, if date is less than so and so copy set of files and if after use this set. Rather once these emissions set is identified there is just one set of yaml files that are NOT date dependent, unlike the old ExtData.
3. Once the proper yaml files are copied (and once again this is NOT date dependent). There is a little script that gets run that examines the GEOS_ChemGridComp.rc and constructs the main extdata.yaml file based on what is actually turned on. So the DAS scripting will need to run this before executing the model:

```
$GEOSBIN/construct_extdata_yaml_list.py GEOS_ChemGridComp.rc
```
and that should be it, much simper really, no more magic dates.

### LINKED ISSUE #345: runjob.pl and checkinput.pl updates
Update the runjob and checkinput perl scripts to set $ESMABIN equal to their directory location rather than using a hard-coded value. This will allow the scripts to work correctly if the build's install directory location is changed.

An option was added to the runjob script to call checkinput. This gives more visibility to the checkinput script.

### LINKED ISSUE #378: inconsistent file name for das.aod in atmens
gcm_ensset_rc.csh ( line#283)  link aod analysis file to  das.aod_a.sfc.%y4%m2%d2_%h2%n2z.mem001.nc4 . But in GAAS_GridComp_ExtData.yaml,  the template is : das.aod_a.sfc.%y4%m2%d2_%h2%n2z.nc4.  
This inconsistency results in that ens gcm do not  use aod analysis even when GAAS is on. 
When MAPL ExtData  routines uses GAAS yaml information, it cannot find aod files in the work dir.  Up to MAPL 2.52 , MAPL ExtData routines handle it with warning if aod files not found,  gcm will run without das.aod input.   Since MAPL 2.55  if idas.aod files ( yaml template ) are not found,  MAPL ExtData routines  crash, ens gcm will not run.

rtodling:
These are changed needed to run GEOS-JEDI when fv3-jedi PR 1335 makes into develop.

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -92,7 +92,7 @@ GEOSana_GridComp:
 mksi:
    local: ./src/Components/@GEOSana_GridComp/GEOSaana_GridComp/GSI_GridComp/@mksi
    remote: ../GEOS_mksi.git
-   tag: v5.42.6
+   tag: v5.42.7
    develop: develop
 
 GEOSgcm_GridComp:
@@ -251,7 +251,7 @@ umwm:
 GEOSgcm_App:
   local: ./src/Applications/@GEOSgcm_App
   remote: ../GEOSgcm_App.git
-  tag: v2.3.8.4
+  tag: v2.3.8.5
   develop: develop
 
 Ocean-LETKF:
--- a/src/Applications/GEOSdas_App/jedi/etc/CMakeLists.txt
+++ b/src/Applications/GEOSdas_App/jedi/etc/CMakeLists.txt
@@ -12,6 +12,7 @@ set (ALLETC
    jedi_acquire_ebkg.j
    jedi_acquire_ioda.j
    jedi_acquire_vbc.j
+   jedi_diffstates.j
    jedi_run_var.j
    mkiau.rc.tenv
    obsop_name_map.yaml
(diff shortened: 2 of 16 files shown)


rtodling:
Among other things, this addresses issue #378

rtodling:
For consistency, I will point out that this has also been addressed (indeed since 5.30x): issue #232

rtodling:
I believe issue #345 has also been dealt w/ sometime ago.

mathomp4 (approved):

mathomp4:
@rtodling I labeled as 0-diff if only to let the label enforcer be happy. Might be might not be. The change is a bit beyond me :)

PR #374, opened 2025-05-21
Fix for incorrect handling of AOD in fcst mode

rtodling:
This is zero-diff in ADAS cycling; but non-zero diff in offline forecasts. 

This fixes a bug in offline forecasts, which have not been seeing the AOD analyses, likely, since the introduction of G2G. Rob Lucchesi notice something odd in the gcm.log in the forecast and an inconsistency w/ when the DAS runs ... I looked closely into it and found that indeed the forecasts were not creating the proper links to the AOD files so the model would see them.

There is also an issue related to the bundleParser.py that has not been working properly in either ADAS or ADAS-fcsts. It's not clear why things related to the setting of Chem still work. The fix brought here related to this, does not seem to affect results in one instance testing - not clear whether results change when a year boundary is crossed, or something else. From what I can tell, the failure of the program to run should have been causing a crash ... but it's not!

I am indicating zero-diff since the bug here affects only offline forecasts which are not done in FP.

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -92,7 +92,7 @@ GEOSana_GridComp:
 mksi:
    local: ./src/Components/@GEOSana_GridComp/GEOSaana_GridComp/GSI_GridComp/@mksi
    remote: ../GEOS_mksi.git
-   tag: v5.42.4
+   tag: v5.42.5
    develop: develop
 
 GEOSgcm_GridComp:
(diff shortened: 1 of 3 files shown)


rtodling:
This also now points to a revised version of gcm_App that fixes aerosol related output used in FP/FPP.

mathomp4 (approved):

PR #392, opened 2025-07-28
Changes in GEOSadas repo to run monthly means jobs with linked inputs

### LINKED ISSUE #382: Month Means: No need to fetch
I'd like to ask that we implement an option - which should become the new default - where the monthly means get calculated directly on the "archived" data as opposed to having a fetch step in advance of the calculation that pulls the data out onto another disc and then perform the calculation.

The present implementation, with the fetch step, was motivated during a time when the archive was a separate entity, when local disc within the archive was limited (after data was brought from tape down to disc), and mpi programs were not supposed to run off of the deporary  holding disc. So the data had to be physically copied to a no-backup area so the calculation could take place.

Presently, with the old archiving system gone, all disc is local, so we should be able to do the calculation without having to copy the data anywhere else. 

This issue request this no-fecth option to be implemented, as an option, that is, preserving still the currently ability to fetch the data and place the data in an alternative disc. The reason why I ask for us to preserve the present fetch ability is that I don't know how things will work when/if we do things on the cloud. 

The no-fetch capability will also spare user space, avoiding temporary duplication of data.

gmao-jstassi:
* monthly.yyyymm.pl.tmpl
  - added -prefetch flag to prefetch inputs as a separate job
  - modified to tar automatically with -means flag, unless -notar flag is given
  - tightened the logic for setting dependFLG for job submittal
  - remove work directory after last means job, if there are no tarandclean jobs
  - changes necessary for tarandclean jobs to do fetch if necessary
  - added sub dump_JOBID to help debugging job dependencies
* monthly_means.j.tmpl
  - added tarFLG to write, during fetch, files needed for tarring
  - added lastFLG to tell means job to clean workdir if no tarandclean job follows
* monthly_means.pl
  - added links flag when calling rget to link inputs if available locally
  - moved final file check to more central location
* monthly_prefetch.j.tmpl
  - removed links flag for separate prefetch jobs; these inputs will be copied, not linked
* monthly_tarandclean.j.tmpl
  - added code to fetch inputs if not already fetched by fetch or means job
  - added ‘h’ flag to tar commands to work with linked files
* testsuites/runjob.pl
  - change “Check for previous use of expid” default from “y” to “n”

Closes #382

~Contigent on PR #3 in GMAO_perllib repo.~ GMAO_perllib v1.2.0 released.

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -61,7 +61,7 @@ GEOS_Util:
 GMAO_perllib:
   local: ./src/Shared/@GMAO_Shared/@GMAO_perllib
   remote: ../GMAO_perllib.git
-  tag: v1.1.0
+  tag: v1.2.0
   develop: main
 
 # When updating the MAPL version, also update the MAPL version in the
--- a/src/Applications/GEOSdas_App/monthly.yyyymm.pl.tmpl
+++ b/src/Applications/GEOSdas_App/monthly.yyyymm.pl.tmpl
@@ -14,12 +14,18 @@
 #    - change file permissions to be executable
 #
 # 2. runtime processing flags:
+#    -prefetch: prefetch inputs before running monthly means (obsolescent)
 #    -means: run monthly means
 #    -plots: run monthly plots
 #    -tar: tar monthly inputs
 #    (see usage() for more complete processing information)
 #
-# 3. To reiterate from Note #1:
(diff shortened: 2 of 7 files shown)


mathomp4:
@rtodling Now that @gmao-jstassi has updated the `components.yaml` for an updated GMAO_perllib this is a self-contained PR. You can test as desired

PR #398, opened 2026-01-15
Feature/jemccurr/hdraob preperror

jemccurr:
Including change to the prepobs error table relevant to high-res sonde assume. The new error table replicates errors for type 120/220 tables with all new high-res types.


rtodling (approved):
Looks fine by me.

PR #399, opened 2026-01-15
Feature/pcastell/viirs updates

patricia-nasa:
This copies over the nnr scripts from GEOSadas and puts them in the AeroML repository under Applications/GAAS_App/
NNR scripts are now pulled into GEOSadas from AeroML.
This is also in preparation of adding VIIRS.

This PR also updates AerosolAnalysis.sh and fvsetup to include VIIRS NOAA-20.


rtodling:
@patricia-nasa @vbuchard Patricia, I am a little confused. We talked about incorporating these VIIRS changes as zero-diff - as long as - VIIRS data are not used. Is this stil the case?

rtodling:
Notice: the version of python used in this version is such that it introduces a non-zero diff in the NNR, including when only using the thus-far MODIS observations. 

A exercise of zero-diff can be satisfied when using only AERONET in say v5.42.12 and v5.43.0. This has been done when putting together the 5.43 version of ADAS and has been shown to produce zero-diffs.

rtodling (commented):
This all seems to work now and it will make up 5.43.0.

@patricia-nasa and @vbuchard Thank you so much for all the work.

rtodling (approved):
This all seems to work now and it will make up 5.43.0.

@patricia-nasa and @vbuchard Thank you so much for all the work.

PR #403, opened 2026-01-21
added obserror for VN20DBL and VN20DBO to avoid an I90/lablin in the log

vbuchard:
Updated PSAS.rc to avoid seeing an I90/lablin in the log for VN20DBL amd VN20DBO.

 **No impact** on the VIIRS DA since those VIIRS data are not currently used.


rtodling (approved):
Thank you Virginie.

PR #410, opened 2026-01-30
Fix for APert 5.43 ESMF consistency

rtodling:
Fix for APERT in light of MAPL changes in 5.43.


mathomp4 (approved):

mathomp4:
Wait. Not sure if zero-diff. I'll let @rtodling add that!

rtodling:
Ah thank you.

rtodling:
What I mean by zero diff here is that the DAS itselft is zero diff - I don't count the pert results on that - the diffs are minor and don't affect the cycle (they are diagnostic stuff)

PR #414, opened 2026-02-17
Feature/rtodling/prep for x54

rtodling:
These a upgrades to turn on the following but default:

- Revised QC for IR
- Revised QC for satwinds
- Use of CrIS-N21
- Use of ABI

This also includes latest update for high-resolution RAOB, but those are not on by default.

ZERO DIFF TO PRIOR: No (intentionally)


rtodling:
@mathomp4 the auto msg above is odd to say the least. I have added a label  (non-zero diff) and it's still asking for a label, why would that  be?

mathomp4 (approved):

PR #423, opened 2026-03-12
no check for automation

weiyuan-jiang:
This change is for nocheck automation.


weiyuan-jiang:
@rtodling , we are trying to test GEOSadas every time before we release MAPL. Now we can run the test case C90C.input with @mathomp4 's restart. What are the output files do you think are important and we should compare with?

rtodling (approved):
fine by me. Thank you for this.

PR #428, opened 2026-04-28
replace link to jpp x51 experiment with user $FVHOME/run/jedi

karpob:
Minor fix to GEOS-JEDI to point to the user's configuration instead of x51/jprepp experiment. Currently will bring up a permissions error when trying to access obsop_name_map.yaml.


rtodling (approved):
looks good by me

PR #429, opened 2026-04-29
Minor adjustments to ABI and fix in ncdiag output

rtodling:
fix for setuprad duble line in ncdiag write (upsetting the contents in the file)
fix for storage of abi-related output.

This also introduces ability to use new linkbcs programs - but not default at this point. 

zero-diff to prior.


mathomp4 (approved):

mathomp4:
Not sure why the label enforcer is acting weird. Must be a github actions issue?

PR #439, opened 2026-06-30
Feature/rtodling/geos jedi 543 to 544

rtodling:
This mainly splots the jedi observing system in a way similar to how SWELL has it. This not only bring GEOS-JEDI to be more in sync w/ SWELL, but gives GEOS-JEDI more flexibility in handling the observing system.

This also implements on-the-fly conversion of GSI ncdiags to IODA. 

Alll changes are zero diff for the GSI-related sytems, ie, GEOS.

Also updates version of g5pert - minor change, zero diff.


mathomp4 (approved):
33000 line change? Trust in @rtodling !

PR #441, opened 2026-07-07
Updates for M21C to use a new option for precip correction

sdrabenh:
Commits GEOSadas resource file updates to accommodate new option of precip correction using scaled climatology and includes recent observing system mods. Note this change is both 0-diff and non-zero-diff depending on the time range. 

This PR is marked with DNA for the moment as we await tags for the following:
- [x] https://github.com/GEOS-ESM/GEOSana_GridComp/pull/238
- [x] https://github.com/GEOS-ESM/GMAO_Shared/pull/452
- [x] https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1468
- [x] https://github.com/GEOS-ESM/GEOSgcm_App/pull/894

Once those tags have been issues, `components.yaml` must be updated with those new tags

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -28,7 +28,7 @@ NCEP_Shared:
 GMAO_Shared:
   local: ./src/Shared/@GMAO_Shared
   remote: ../GMAO_Shared.git
-  tag: v1.7.2+R21C_v1.0.5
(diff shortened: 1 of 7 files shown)


rtodling (approved):
Looks to me you've done the work.

